PR #24, opened 2023-10-19
ATB-1113 created lambda to invoke private apigateway

KG700:
## Proposed changes

[ATB-1113 | Create lambda to invoke private apigate](https://govukverify.atlassian.net/browse/ATB-1113)

### What changed

Created a lambda that invokes the private api gateway in the dev environment.

### Why did it change

This enables developers to test the functionality without the need to create a public facing api.

Currently a DRAFT PR because:
- Requires this ticket to be merged first so the PrivateApiGateway is in the template: [ATB-1080](https://govukverify.atlassian.net/browse/ATB-1080)

## Checklists


liamquayle (approved):

MinaMahdi (approved):

jessica-li-01 (approved):

src/services/app-config-service.ts
```diff
@@ -41,6 +41,28 @@ export class AppConfigService {
...
+
+  public get queryParameters() {
+    return process.env['QUERY_PARAMETERS'];
+  }
+
+  public get httpRequestMethod() {
+    return this.validateConfiguration('HTTP_REQUEST_METHOD');
+  }
+
   /**
```
KG700, line 66: Do we want this lambda to use the app config service to access the environment variables? Or would it be better to keep this lambda completely separate from the main solution?

src/handlers/invoke-private-api.ts
```diff
@@ -0,0 +1,52 @@
...
+  queryParameters?: string;
+  headers?: {
+    [key: string]: string;
+  };
+}
+
+const appConfig = AppConfigService.getInstance();
+
+export const handle = async (event: CustomEvent) => {
+  const userId = event.userId ?? appConfig.userId;
```
chiarapascucci-gds, line 15: is there a reason why the userId can be set as a environment variable? The default value through the env var is an empty string, which wouldn't really work as a default value. Would it not be safer to enforce that a userid is passed in the event?

src/infra/main/template.yaml
```diff
@@ -699,6 +700,35 @@ Resources:
...
       Role: !GetAtt StatusRetrieverFunctionRole.Arn
+
+  InvokePrivateAPIGatewayFunction:
+    Condition: IsInternal
+    Type: AWS::Serverless::Function
+    # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
+    # checkov:skip=CKV_AWS_115: Function is not configured for function-level concurrent execution Limit
+    Properties:
+      FunctionName: !Sub '${AWS::StackName}-InvokePrivateAPIGatewayFunction'
+      CodeUri: ../../../dist/handlers/invoke-private-api
```
chiarapascucci-gds, line 711: CodeSigningConfigArn: !If [ UseCodeSigning, !Ref CodeSigningConfigArn, !Ref AWS::NoValue ]

src/handlers/invoke-private-api.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (Object.keys(response).length === 0) {
+    return {
+      statusCode: response.status || 400,
+      message: "Request didn't return a valid response",
+    };
+  }
+
+  return {
+    statusCode: response.status,
+    message: JSON.stringify(response),
```
chiarapascucci-gds, line 50: I would call this field `payload` or `data`? as it will likely be a JSON object with all the data, rather than a message

src/services/app-config-service.ts
TaherKapasi: separate imho

src/handlers/invoke-private-api.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const appConfig = AppConfigService.getInstance();
+
+export const handle = async (event: CustomEvent) => {
+  const userId = event.userId ?? appConfig.userId;
+  let queryParameters = event.queryParameters ?? appConfig.queryParameters;
+  queryParameters = queryParameters ? '?' + queryParameters : '';
+
+  if (!userId) {
```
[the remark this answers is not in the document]
KG700: This is here to catch the case where the userId has not been provided in either the event or the environment variable. It's fine for userId to be undefined in the event if the person using the lambda is using the environment variables to update the userId

src/handlers/invoke-private-api.ts
KG700: Yeah, we wanted it to be possible to use this lambda using either the lambda event or the environment variables.

chiarapascucci-gds: Okay.. I think the env var make sense for things like the endpoint, which are not going to change often. But I really don't understand including the userId and queryParameters in the env vars, sorry - is this how this lambda is going to be used in the test framework?

chiarapascucci-gds (approved):

src/services/account-states/account-state-engine.ts
```diff
@@ -165,8 +165,8 @@ function buildStateTransitionError(metricName: MetricNames, errorMessage: string
 
 /**
  * Helper method to build a ConfigurationError and log relevant information when this type of error is to be thrown
- * @param metricName
- * @param errorMessage
+ * @param metricName -
+ * @param errorMessage -
```
amd-gds, line 169: let's add some description after "-"

amd-gds (approved):

PR #34, opened 2023-10-26
ATB-1079 | Account-Deletion-Processor-Handler

jessica-li-01:
### What changed

- Added lambda for updating status of the table

### Why did it change

- To update the status when an account is deleted

### Issue tracking

- [ATB-1079](https://govukverify.atlassian.net/browse/ATB-1079)

## Testing

1. Updating the DDB with a user_id that hasn't had their status confirmed + Sending the notification via SNS topic of the user_id that has been deleted
Screenshot 2023-10-30 at 15 34 49

2. The status updating to the account being deleted
Screenshot 2023-10-30 at 15 34 57

3. The account of the user being updated to being deleted
Screenshot 2023-10-30 at 15 38 12

4. Coverage Report
Screenshot 2023-10-30 at 16 05 37

## Checklists

[ATB-1079]: https://govukverify.atlassian.net/browse/ATB-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/account-deletion-processor-handler.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+  return userId.trim();
+}
+
+async function updateDeleteStatusId(userId: string) {
+  try {
+    await ddbService.updateDeleteStatus(userId);
+    logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
+  } catch (error) {
+    logger.error(`${LOGS_PREFIX_SENSITIVE_INFO} Error updating account ${userId}`, { error });
```
chiarapascucci-gds, line 59: Is there a reason why we are not throwing the error? Are we not going to re-try the message?

src/services/test/dynamo-db-service.test.ts
amd-gds, line 1: Hey Jess :) I have a couple of comments on this file.
1. Can you add some UTs to test the`putItemForUserId` method works as expected?
2. The `retrieveRecordsByUserId` method will change quite a bit with the PR Chiara and I have, so I would not add all these UTs to the file as not relevant to the updated method

jessica-li-01: Hey Ana, thanks!

@chiarapascucci-gds - shall I add the putItemForUserId UT, as you mentioned that would also change a lot? Let me know

I've deleted all references now I think 😅

chiarapascucci-gds: in our PR so far, the `putItemForUserId` is not there anymore, it will be replaced by the `updateUserStatus` method, so I wouldn't worry about writing unit tests for it now.. We will however maitain the `retrieveRecordsByUserId` method to retrieve the item from the db, but we adding projection expression which would need to be considered in the UTs. So I would say that @amd-gds and I's PR needs to be merged before any further changes to the UTs.

src/handlers/account-deletion-processor-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+function getUserId(record: SQSRecord) {
+  let messageBody: SNSMessage;
+  let message: DeleteStatusUpdateSNSMessage;
+  try {
+    messageBody = JSON.parse(record.body);
+    message = JSON.parse(messageBody.Message);
+  } catch {
+    logger.error('The SQS message can not be accessed.');
```
chiarapascucci-gds, line 32: minor thing: can you please change this error message to something related to the fact that the message cannot per parsed? Saying access may make people think of a permission issue

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
+      Policies:
+        - PolicyName: !Sub "${AWS::StackName}-AccountDeletionProcessorFunctionRolePolicy"
```
chiarapascucci-gds, line 286: can we please have all the policy statements in the policy resource unless there is a reason why those need to be attached to the role?

chiarapascucci-gds: cc. @domantas-gds

src/infra/main/template.yaml
```diff
@@ -177,6 +182,19 @@ Resources:
...
+            Action:
+              - "kms:Decrypt"
+              - "kms:GenerateDataKey"
+            Resource: "*"
+            Condition:
+              ArnEquals:
+                aws:SourceArn: !If
+                  - IsInternal
+                  - !Ref AccountDeletionProcessorSNSTopic
+                  - !FindInMap [ EnvConfig, !Ref Environment, AccountDeletionProcessorSNSTopicARN ]
```
chiarapascucci-gds, line 197: I think this will make the key inaccessible by sns given that we are passing empty strings for any other env but dev.. is this a good idea? @domantas-gds

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
 
+  AccountDeletionProcessorFunction:
+    Type: AWS::Serverless::Function
+    # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
+    # checkov:skip=CKV_AWS_115: Function is not configured for function-level concurrent execution Limit
+    Properties:
+      FunctionName: !Sub '${AWS::StackName}-AccountDeletionProcessorFunction'
+      Description: "A lambda function that updates an account status indicating that the account has been deleted"
+      CodeUri: ../../../dist/handlers/account-deletion-processor-handler
+      Timeout: 10
```
chiarapascucci-gds, line 226: Is there a point in having the visibility timeout value on the queue as high as 60, but have the lambda timeout at 10?

domantas-gds: Nope agreed that shouldn't work? Has this been tested @jessica-li-01

domantas-gds:
Can we link ticket in issue tracking please

src/infra/main/template.yaml
domantas-gds: Yeah agreed :)

jessica-li-01: @domantas-gds @chiarapascucci-gds I had the same issue with VC Storage, these policies have to be inline with the lambda otherwise it wouldn't work, the SNS to SQS queue link could not be established unless it was like this when I checked.

src/infra/main/template.yaml
jessica-li-01: Changed- thanks! :)

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+      Description: "A lambda function that updates an account status indicating that the account has been deleted"
+      CodeUri: ../../../dist/handlers/account-deletion-processor-handler
+      Timeout: 60
+      Handler: account-deletion-processor-handler.handler
+      Role: !GetAtt AccountDeletionProcessorFunctionRole.Arn
+      Environment:
+        Variables:
+          TABLE_NAME:
+            Fn::ImportValue: !Sub "${CoreStackName}-AccountStatusTableName"
+          DELETED_ACCOUNT_STATUS_RETENTION_SECONDS: 63072000
```
domantas-gds, line 233: Can we just leave comment stating this is 2 years please

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+          METRIC_SERVICE_NAME: !Sub "${AWS::StackName}-AccountDeletionProcessorFunction"
+      Events:
+        SNSEvent:
+          Type: SNS
+          Properties:
+            RedrivePolicy: !Sub "{ \"deadLetterTargetArn\": \"arn:aws:sqs:${AWS::Region}:${AWS::AccountId}:${AWS::StackName}-AccountDeletionProcessorDLQ\" }"
```
domantas-gds, line 239: Should be able to dynamically reference the DLQ ARN as I don't see a reference from DLQ to this function. So therefore there wouldn't be a circular dependency

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+              - "sqs:SendMessage"
+            Resource: !GetAtt AccountDeletionProcessorQueue.Arn
+            Principal:
+              Service: sns.amazonaws.com
+            Condition:
+              ArnEquals:
+                aws:SourceArn:
+                  - !If
+                    - IsInternal
+                    - !Sub "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:${AWS::StackName}-AccountDeletionProcessorSNSTopic"
```
domantas-gds, line 378: Could we not dynamically fetch the ARN? Unsure but doesn't look like it would cause a circular dependency

src/infra/main/template.yaml
```diff
@@ -48,19 +48,25 @@ Parameters:
 
 Conditions:
   UsePermissionsBoundary: !Not [ !Equals [ !Ref PermissionsBoundary, 'none' ] ]
+  IsInternal: !Or [ !Equals [!Ref Environment, dev], !Equals [!Ref Environment, build],!Equals [ !Ref Environment, staging]]
 
 Mappings:
   EnvConfig:
     dev:
       LambdaLogLevel: DEBUG
+      AccountDeletionProcessorSNSTopicARN: ""
```
domantas-gds, line 57: If we leave comments next to dev and build to state empty string as pointing to our internal AccountDeletionProcessor SNS topic

src/infra/main/template.yaml
domantas-gds: Yes as discussed let's Ref the AccountDeletionProcessorSNSTopic we've defined until we can populate those mappings :) So will also need to remove the IsInternal condition from the AccountDeletionProcessorSNSTopic resource

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+              Service:
+                - "sns.amazonaws.com"
+            Action:
+              - "kms:Decrypt"
+              - "kms:GenerateDataKey"
+            Resource: "*"
+            Condition:
+              ArnEquals:
+                aws:SourceArn:
+                  - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-AccountDeletionProcessorFunction"
```
domantas-gds, line 448: This policy isn't quite right. Principal service is SNS yet SourceArn points to a Lambda which is a separate service

src/infra/main/template.yaml
```diff
@@ -197,6 +215,258 @@ Resources:
...
+
+  AccountDeletionProcessorResourcePolicy:
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref AccountDeletionProcessorFunction
+      Action: lambda:InvokeFunction
+      Principal: sns.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
+      SourceArn: !If
+        - IsInternal
```
domantas-gds, line 267: If we just Ref our SNS topic until we have values for mappings )

jest.config.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
 process.env['CLOUDWATCH_METRICS_NAMESPACE'] = 'test_namespace';
 process.env['METRIC_SERVICE_NAME'] = 'test';
 process.env['TABLE_NAME'] = 'table_name';
-process.env['AWS_REGION'] = 'aws_region'
+process.env['AWS_REGION'] = 'aws_region';
+process.env['DELETED_ACCOUNT_STATUS_RETENTION_SECONDS'] = '12345';
```
TaherKapasi, line 14: DELETED_ACCOUNT_RETENTION_SECONDS - easier to understand purpose

package.json
```diff
@@ -27,14 +27,19 @@
...
   "dependencies": {
+    "@aws-lambda-powertools/commons": "^1.14.0",
     "@aws-lambda-powertools/logger": "^1.14.0",
-    "@aws-sdk/client-dynamodb": "^3.430.0",
     "@aws-lambda-powertools/metrics": "^1.14.0",
-    "@aws-lambda-powertools/tracer": "^1.14.0"
+    "@aws-lambda-powertools/tracer": "^1.14.0",
+    "@aws-sdk/client-dynamodb": "^3.430.0",
+    "@aws-sdk/util-dynamodb": "^3.433.0",
+    "aws-sdk-client-mock": "^3.0.0",
```
TaherKapasi, line 36: these should be devDependencies

src/handlers/account-deletion-processor-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+import { AppConfigService } from '../services/app-config-service';
+import type { Context, SNSMessage, SQSEvent, SQSRecord } from 'aws-lambda';
+import { DeleteStatusUpdateSNSMessage } from '../data-types/interfaces';
+
+const appConfig = AppConfigService.getInstance();
+const ddbService = new DynamoDatabaseService(appConfig.tableName);
+
+export const handler = async (event: SQSEvent, context: Context): Promise<void> => {
+  logger.addContext(context);
+  if (!event.Records[0] || event.Records.length === 0) {
```
TaherKapasi, line 13: just !event.Records should do

src/infra/main/template.yaml
```diff
@@ -197,11 +217,263 @@ Resources:
...
 
+  AccountDeletionProcessorFunction:
+    Type: AWS::Serverless::Function
+    # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
+    # checkov:skip=CKV_AWS_115: Function is not configured for function-level concurrent execution Limit
+    Properties:
+      FunctionName: !Sub '${AWS::StackName}-AccountDeletionProcessorFunction'
+      Description: "A lambda function that updates an account status indicating that the account has been deleted"
+      CodeUri: ../../../dist/handlers/account-deletion-processor-handler
+      Timeout: 60
```
TaherKapasi, line 228: should reduce this - will never take 60 secds to run - lets set it to 10

src/infra/alarm/template.yaml
```diff
@@ -5,10 +5,6 @@ Description: >-
   Alarm template for the Account Interventions Service (AIS) solution.
 
 Parameters:
-  CoreStackName:
```
TaherKapasi, line 8: is this not needed in the template at ln 234

src/infra/main/template.yaml
```diff
@@ -197,11 +217,263 @@ Resources:
...
+      AliasName: !Sub "alias/${AWS::StackName}-AccountDeletionProcessorSNSKmsKey"
+      TargetKeyId: !GetAtt AccountDeletionProcessorSNSKmsKey.Arn
+
+ #### TxMA ####
   TxMAIngressQueue:
     Type: AWS::SQS::Queue
     Properties:
       QueueName: !Sub "${AWS::StackName}-TxMAIngressQueue"
-      MessageRetentionPeriod: 1209600
+      MessageRetentionPeriod: 1209600 #Two Years in seconds
```
TaherKapasi, line 476: two years in secs is 63072000

src/handlers/account-deletion-processor-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { Context, SNSMessage, SQSEvent, SQSRecord } from 'aws-lambda';
+import { DeleteStatusUpdateSNSMessage } from '../data-types/interfaces';
+
+const appConfig = AppConfigService.getInstance();
+const ddbService = new DynamoDatabaseService(appConfig.tableName);
+
+export const handler = async (event: SQSEvent, context: Context): Promise<void> => {
+  logger.addContext(context);
+  if (!event.Records[0]) {
+    logger.error('The event does not contain any records.');
```
liamquayle, line 14: Going to ask what is probably a stupid question so apols, but if the event gets passed in with no records, do we not want to send a metric off to cloudwatch due to an erroneous SQSEvent and throw an error here rather than later on? Or is none of that needed?

src/infra/main/template.yaml
```diff
@@ -152,21 +158,81 @@ Resources:
       Roles:
         - !Ref AISSampleLambdaHandlerRole
 
- #### TxMA ####
+  AccountDeletionProcessorSNSTopic:
```
TaherKapasi, line 161: can we add a condition on this so that the topic is created in dev and build envs only, IsInternal possibly

src/services/dynamo-db-service.ts
```diff
@@ -103,4 +89,25 @@ export class DynamoDbService {
...
+      UpdateExpression: 'SET #isAccountDeleted = :isAccountDeleted, #ttl = :ttl',
+      ExpressionAttributeNames: {
+        '#isAccountDeleted': 'isAccountDeleted',
+        '#ttl': 'ttl',
+      },
+      ExpressionAttributeValues: {
+        ':isAccountDeleted': { BOOL: true },
+        ':ttl': { N: ttl.toString() },
+      },
+      ConditionExpression: 'attribute_exists(pk)',
```
TaherKapasi, line 107: lets add a condition that its only updated if `isAccountDeleted` is not present or is `false` otherwise the update fails but the message is ignored and redriven back to the queue

TaherKapasi (changes requested):

src/infra/main/template.yaml
chiarapascucci-gds: Hey @TaherKapasi , @domantas-gds and I discussed this with Jess yesterday. My understanding is that we do not have the ARN of topics that we need yet, so for now we have plugged this placeholder topic in all environments, so that deployment wouldn't fail if and when we go past dev/build (as we were using empty strings). cc. @domantas-gds 
The idea is to go back to the condition once we have arns for the higher envs

jessica-li-01: Thanks Chiara!

src/handlers/account-deletion-processor-handler.ts
jessica-li-01: I think we'll add metrics later on in a separate ticket if necessary :)

TaherKapasi (approved):

liamquayle (approved):

PR #44, opened 2023-11-02
ATB-1136 | Refactor account-deletion-processor

jessica-li-01:
### What changed

- Refactor/ change the condition expression
- Re-driving in the queue based on condition expression
- Updated UTs

### Why did it change

- Make sure that a user is not re-driven in the queue when the user does not exist (ConditionalCheckFailedException)
- Make sure that the users are re-driven if the error is anything but a Conditional Check Exception.

### Issue tracking

- [ATB-1136](https://govukverify.atlassian.net/browse/ATB-1136)

## Testing

Scenario 1: When the user_id does not exist in the table, and an error is logged when the user_id is trying to be deleted.
Screenshot 2023-11-02 at 15 27 13
Screenshot 2023-11-02 at 15 26 56

Scenario 2: Adding user_id, deleting user_id, marks as deleted correctly.

Screenshot 2023-11-02 at 15 29 13
Screenshot 2023-11-02 at 15 28 54
Screenshot 2023-11-02 at 15 29 36
Screenshot 2023-11-02 at 15 29 59

UTs:
Screenshot 2023-11-02 at 15 47 20

Logs (re; ConditionalCheckFailedException)
If it isn't a conditional check exception, the SQS record will be sent to the queue to be retried:
Screenshot 2023-11-07 at 14 54 17

When the user does not exist, a log line is logged, and the error is not retried:
Screenshot 2023-11-07 at 14 47 21

## Checklists

[ATB-1136]: https://govukverify.atlassian.net/browse/ATB-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/account-deletion-processor-handler.ts
+++ b/src/handlers/account-deletion-processor-handler.ts
@@ -55,8 +55,6 @@ function getUserId(record: SQSRecord) {
 async function updateDeleteStatusId(userId: string) {
   try {
     await ddbService.updateDeleteStatus(userId);
-    logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
-    logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
   } catch (error) {
     logger.error(`${LOGS_PREFIX_SENSITIVE_INFO} Error updating account ${userId}`, { error });
     logAndPublishMetric(MetricNames.MARK_AS_DELETED_FAILED);
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -130,7 +130,8 @@ export class DynamoDatabaseService {
         ':ttl': { N: ttl.toString() },
         ':false': { BOOL: false },
       },
-      ConditionExpression: 'attribute_not_exists(isAccountDeleted) OR isAccountDeleted = :false',
+      ConditionExpression:
+        'attribute_exists(pk) AND (attribute_not_exists(isAccountDeleted) OR isAccountDeleted = :false)',
     };
     const command = new UpdateItemCommand(commandInput);
     const response = await this.dynamoClient.send(command);
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -136,6 +136,8 @@ export class DynamoDatabaseService {
     const command = new UpdateItemCommand(commandInput);
     try {
       const response = await this.dynamoClient.send(command);
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
       if (!response) {
         const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
         logger.error(errorMessage);
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -136,7 +136,7 @@ export class DynamoDatabaseService {
     const command = new UpdateItemCommand(commandInput);
     try {
       const response = await this.dynamoClient.send(command);
-      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted.`);
       logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
       if (!response) {
         const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
@@ -145,7 +145,7 @@ export class DynamoDatabaseService {
       }
       return response;
     } catch {
-      const errorMessage = `${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}`;
+      const errorMessage = `${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}.`;
       logger.error(errorMessage);
       logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
     }
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -145,7 +145,11 @@ export class DynamoDatabaseService {
       }
       logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
       return response;
-    } catch {
+    } catch (error: any) {
+      console.log(error);
+      if (error.name === 'ValidationException') {
+        throw new Error();
+      }
       const errorMessage = `${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}.`;
       logger.error(errorMessage);
       logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -134,12 +134,18 @@ export class DynamoDatabaseService {
         'attribute_exists(pk) AND (attribute_not_exists(isAccountDeleted) OR isAccountDeleted = :false)',
     };
     const command = new UpdateItemCommand(commandInput);
-    const response = await this.dynamoClient.send(command);
-    if (!response) {
-      const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+    try {
+      const response = await this.dynamoClient.send(command);
+      if (!response) {
+        const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+      }
+      return response;
+    } catch {
+      const errorMessage = `Error updating item with pk ${userId}`;
       logger.error(errorMessage);
       logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
     }
-    return response;
   }
 }
--- a/src/services/dynamo-database-service.ts
+++ b/src/services/dynamo-database-service.ts
@@ -14,9 +14,9 @@ import { NodeHttpHandler } from '@smithy/node-http-handler';
 import tracer from '../commons/tracer';
 import { unmarshall } from '@aws-sdk/util-dynamodb';
 import { logAndPublishMetric } from '../commons/metrics';
-import { TooManyRecordsError } from '../data-types/errors';
 import { getCurrentTimestamp } from '../commons/get-current-timestamp';
 import { DynamoDBStateResult } from '../data-types/interfaces';
+import { TooManyRecordsError } from '../data-types/errors';
 
 const appConfig = AppConfigService.getInstance();
 export class DynamoDatabaseService {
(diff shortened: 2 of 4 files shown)


amd-gds (approved):

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,24 @@ export class DynamoDatabaseService {
...
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
+      if (!response) {
+        const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+      }
+      return response;
+    } catch {
+      const errorMessage = `Error updating item with pk ${userId}`;
```
chiarapascucci-gds, line 118: I think you should add the sensitive info prefix here since we are logging the user id

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,24 @@ export class DynamoDatabaseService {
...
+        'attribute_exists(pk) AND (attribute_not_exists(isAccountDeleted) OR isAccountDeleted = :false)',
     };
     const command = new UpdateItemCommand(commandInput);
-    const response = await this.dynamoClient.send(command);
-    if (!response) {
-      const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+    try {
+      const response = await this.dynamoClient.send(command);
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted`);
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
```
chiarapascucci-gds, line 110: if the response is undefined, we are still logging this metric but then we are also logging DB UPDATE ERROR... is that intentional? should we not `logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);` only when the operation has definitely succeeded ?

chiarapascucci-gds: also.. what is the response if the update is NOT made because of the condition expression? I do not think it will throw an error if the update is not done because the condition is not met.. so then we may still be logging this success metric when it is not really relevant?

chiarapascucci-gds:
Hey Jess - I know this is a refactoring PR, but the update condition is modified thus the flow is also modified, so I would expect to see a completed PR template with testing done and evidence of that testing. Please complete the check list and link the ticket as well

src/services/dynamo-database-service.ts
jessica-li-01: When I tested it, it didn't log the 'MARK_AS_DELETED_SUCCEEDED' when the user_id didn't exist. If it's undefined it should just log DB_UPDATE_ERROR.

TaherKapasi:
Hi @jessica-li-01 - could you test and provide evidence in the PR with screenshots - thanks - we must do this everytime no matter how small the change as we could get scrutinised over this in the future

TaherKapasi (changes requested):
please see comment on filling out the PR  template

jessica-li-01:
> please see comment on filling out the PR template

Yup, sorry forgot, all done

src/services/dynamo-database-service.ts
chiarapascucci-gds: okay - but as it is written right now, if the an error is not thrown but the response is undefined (which I know it is not expected) then we would first log a success metric and then log a db update error inside the if statement block. I would move the success metric logging just before the response is returned and also add a return statement inside the if block to ensure that we do not log both success and failure metric in one flow

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted.`);
+      if (!response) {
+        const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+        return;
+      }
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
+      return response;
+    } catch (error: any) {
```
TaherKapasi, line 118:
```suggestion
    } catch (error: unknown) {
```

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+      if (error.name != 'ConditionalCheckFailedException') {
```
TaherKapasi, line 119:
```suggestion
      if (!error.name || error.name != 'ConditionalCheckFailedException') {
```

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+        const errorMessage = `${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}.`;
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+        throw new Error('Error was not a Conditional Check Exception.'); //Therefore re-driving message back to the queue.
```
TaherKapasi, line 123:
```suggestion
        throw new Error('Error was not a Conditional Check Exception.', { error }); //Therefore re-driving message back to the queue.
```

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+        const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+        return;
+      }
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
+      return response;
+    } catch (error: any) {
+      if (error.name != 'ConditionalCheckFailedException') {
+        const errorMessage = `${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}.`;
```
TaherKapasi, line 120:
```suggestion
        logger.error(`${LOGS_PREFIX_SENSITIVE_INFO} Error updating item with pk ${userId}.`);
```

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+        logger.error(errorMessage);
```
TaherKapasi, line 121: remove based on line above

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+        throw new Error('Error was not a Conditional Check Exception.'); //Therefore re-driving message back to the queue.
+      }
+      logger.info('No intervention exists for this account.');
```
TaherKapasi, line 125:
```suggestion
      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} No intervention exists for this account.`, { userId });
```

src/services/dynamo-database-service.ts
```diff
@@ -100,15 +100,29 @@ export class DynamoDatabaseService {
...
-      logger.error(errorMessage);
-      logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+    try {
+      const response = await this.dynamoClient.send(command);
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted.`);
+      if (!response) {
+        const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
+        logger.error(errorMessage);
+        logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+        return;
```
TaherKapasi, line 114: should throw an error here as well as the db responded with nothing so we are unsure if the update occurred, therefore redrive.
```suggestion
        throw new Error('DB returned a null response when attempting an Update.');
```

TaherKapasi (changes requested):

src/services/dynamo-database-service.ts
jessica-li-01: Unable to include error as an object here, as the error is unknown/ any and it has to be specified to be used here

src/services/dynamo-database-service.ts
jessica-li-01: Doesn't accept unknown, only any

src/services/dynamo-database-service.ts
jessica-li-01: In the UT response can't be null or undefined, so it's very unlikely to go into the block. I remember Chiara mentioning maybe removing that !response block? Otherwise happy to change it, but I don't seem to be able to test that error in the UT- it won't let me assert a null and doesn't like the undefined as any

src/handlers/account-deletion-processor-handler.ts
```diff
@@ -55,8 +55,6 @@ function getUserId(record: SQSRecord) {
 async function updateDeleteStatusId(userId: string) {
```
liamquayle, line 55: Just as a general comment on this file, could we possibly get code documentation here?

src/services/test/dynamo-database-service.test.ts
```diff
@@ -11,7 +11,7 @@ import { getCurrentTimestamp } from '../../commons/get-current-timestamp';
 import logger from '../../commons/logger';
 import { TooManyRecordsError } from '../../data-types/errors';
 import { logAndPublishMetric } from '../../commons/metrics';
-import { MetricNames } from '../../data-types/constants';
+import { MetricNames} from '../../data-types/constants';
```
liamquayle, line 14: absolute nit pick, sorry! Could we get the space added back in please?

src/services/dynamo-database-service.ts
```diff
@@ -130,15 +130,22 @@ export class DynamoDatabaseService {
...
-      const errorMessage = 'DynamoDB may have failed to update items, returned a null response.';
-      logger.error(errorMessage);
-      logAndPublishMetric(MetricNames.DB_UPDATE_ERROR);
+    try {
+      const response = await this.dynamoClient.send(command);
+      logger.info(`${LOGS_PREFIX_SENSITIVE_INFO} Account ${userId} marked as deleted.`);
+      logAndPublishMetric(MetricNames.MARK_AS_DELETED_SUCCEEDED);
+      return response;
+    } catch (error: any) {
+      if (!error.name || error.name != 'ConditionalCheckFailedException') {
```
chiarapascucci-gds, line 143: change `!=` to `!==`

src/services/dynamo-database-service.ts
chiarapascucci-gds: Hey Jess, I have tried this and you should be able to do:
```
catch (error: any) {
    logger.error("message", { error: error });
    if(!error.name || error.name == "something")
```
so that we do not loose the original error message

src/services/test/dynamo-database-service.test.ts
```diff
@@ -149,20 +149,31 @@ describe('Dynamo DB Service', () => {
...
     const dynamoDBService = new DynamoDatabaseService('table_name')
     await dynamoDBService.updateDeleteStatus('hello');
     expect(ddbMock).toHaveReceivedCommandWith(UpdateItemCommand, commandInput);
   });
 
-  it('should throw an error if response of the Dynamo Client is null / undefined', async () => {
+  it('throws an error when it fails to update the userId status.', async () => {
     const mockedUpdateCommand = mockClient(DynamoDBClient).on(UpdateItemCommand);
-    mockedUpdateCommand.resolves(undefined as any);
+    mockedUpdateCommand.rejectsOnce("InternalServerError" || "InvalidEndpointException" || "ItemCollectionSizeLimitExceededException" || "ProvisionedThroughputExceededException" || "RequestLimitExceeded" || "ResourceNotFoundException" || "TransactionConflictException" || "DynamoDBServiceException" || undefined );
```
chiarapascucci-gds, line 161: why are we providing multiple values here?

jessica-li-01: These are all the errors other than ConditionalCheckFailed. It doesn't let me do it generically I think

PR #49, opened 2023-11-03
ATB-1145 | Code Smell fix and refactoring of State Engine

chiarapascucci-gds:
## Proposed changes
This PR merges changes to refactor the state engine class to decrease cognitive complexity, increase readability and maintainability of code

### What changed
Main changes are all in the State Engine class: 
- class has been refactored to follow the singleton pattern 
- configuration for the class have been changed to an object containing information of state and allowed transitions as a graph data structure: the different account states are the nodes/vertexes in the graph, the edges are all the different transitions that can happen (intervention or user action) and links between nodes through the edges are represented by a set of adjacency lists (which represent the allowed transition from a given state)
- reworded some of the metrics to make it clearer what they are logging
- updated unit tests to cover all configuration errors scenario
- add more complete validation on the configuration object given to the account state engine class, so that in practice a lot of the error scenarios should not be possible / if validation doesn't pass when the class instance is created, an error is thrown and processing does not proceed

### Why did it change
decrease cognitive complexity, increase readability and maintainability of code

### Issue tracking

- [ATB-1145](https://govukverify.atlassian.net/browse/ATB-1145)

## Testing
All possible transition combinations are covered through unit testing
Screenshot 2023-11-06 at 13 04 47

deployed changes to AWS account and tested a few transitions + edge cases
- supended no user action -> suspended id and psw reset -> user action psw reset -> user action id reset -> unsuspended (successful)
- unexpected account state retrieved -> metric is logged and error thrown
- non-existing user -> block account (successful)
Screenshot 2023-11-06 at 13 07 41

Sonarcloud reports no code smells / issues
Screenshot 2023-11-06 at 14 53 28

## Checklists

[ATB-1145]: https://govukverify.atlassian.net/browse/ATB-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TaherKapasi (approved):

src/services/account-states/account-state-engine.ts
```diff
@@ -1,118 +1,175 @@
...
+   * It finds the transition that the event is trying to apply
+   * It checks if this transition can be found in the allowed transition for the state.
+   *  if Yes, it returns a Partial for UpdateItemCommandInput to use in the database command
+   *  if Not, it throws an error
+   * @param event - EventEnum representation of event received
+   * @param currentState - optional state object representation the current state of the account, it defaults to account unsuspended if nothing is passed
+   */
+  applyEventTransition(event: EventsEnum, currentState?: StateDetails) {
+    if (!currentState)
+      currentState = {
```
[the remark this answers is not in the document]
chiarapascucci-gds: I thought about doing this, but the interface does not describe the node object to necessarily have this key, thus I would then need to manage the event in which this specific key is undefined, or add a non-null assertion, which is not totally safe as this key could be missing from the object.

src/services/account-states/account-state-engine.ts
```diff
@@ -1,118 +1,189 @@
...
-      if (!currentState.allowedTransitions.includes(proposedTransition)) {
-        logAndPublishMetric(MetricNames.STATE_TRANSITION_NOT_ALLOWED_OR_IGNORED);
-        throw new StateTransitionError(
-          `Intervention ${proposedTransition} is now allowed on the current account state ${accountStateEventEnum}.`,
-        );
-      }
+  /**
+   * Helper method to return the list of possible transitions given a state name
+   * If none can be found it throws an error
+   * @param nodeKye - the state name used as the key to identify the nodes in the config object
```
KG700, line 108: What does Kye mean? (or is this a typo?)

chiarapascucci-gds: typo ! thanks

src/services/account-states/account-state-engine.ts
```diff
@@ -1,118 +1,189 @@
...
+   */
+  private findPossibleTransitions(nodeKye: string) {
+    const allowedTransition = AccountStateEngine.configuration.adjacency[nodeKye];
+    if (!allowedTransition)
```
KG700, line 112: `allowedTransition` is an array, right? I think the variable name should be plural. And I'm pretty sure `![]` will return false, so you'll probably need to update the if to `allowedTransitions.length` (or something to that effect).

chiarapascucci-gds: thanks - will change the name to plural. But the condition is right, I am checking for when the variable is undefined overall, not array is empty. Empty array is valid, as a state may have no transitions allowed from it :)

src/services/account-states/account-state-engine.ts
```diff
@@ -1,118 +1,189 @@
...
+  return new StateTransitionError(errorMessage);
+}
+
+/**
+ * Helper method to compare two StateDetails objects.
+ * It returns true, IFF all fields are the same, false otherwise
+ * @param aState - first account state
+ * @param anotherState - second account state
+ */
+function compareAccountStates(aState: StateDetails, anotherState: StateDetails) {
```
KG700, line 171: I wonder if changing the function name to something like `isAccountStatesTheSame` (or `areAccountStatesTheSame` would be better english) would make it clearer what the return value means

src/services/account-states/account-state-engine.ts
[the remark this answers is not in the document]
KG700: What I think you could do is create an enum for the transitions. Then update the interface to:
```
export interface TransitionConfigurationInterface {
  nodes: {
    [key in Transitions]: StateDetails;
  };
...
```

src/services/account-states/account-state-engine.ts
```diff
@@ -1,118 +1,189 @@
...
-        return key as EventsEnum;
+  private static validateConfiguration() {
+    const accountStates = Object.keys(AccountStateEngine.configuration.nodes).sort(compareStrings);
+    const adjacencyLists = Object.keys(AccountStateEngine.configuration.adjacency).sort(compareStrings);
+    if (JSON.stringify(adjacencyLists) !== JSON.stringify(accountStates))
+      throw buildError(
+        MetricNames.INVALID_STATE_ENGINE_CONFIGURATION,
+        'Invalid state engine configuration detected. Adjacency mismatch',
+      );
+    for (const element of Object.values(AccountStateEngine.configuration.edges)) {
```
KG700, line 33:
```suggestion
    for (const edge of Object.values(AccountStateEngine.configuration.edges)) {
```
Seems slightly better to me

amd-gds (approved):

src/services/account-states/account-state-engine.ts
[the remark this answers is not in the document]
chiarapascucci-gds: I want to keep things simple, and restrict the amount of Enums, which would mean whenever we add a state we would have extra places where to update the info

PR #89, opened 2023-11-23
ATB-1203 | History Field 

chiarapascucci-gds:
## Proposed changes
[ATB 1203](https://govukverify.atlassian.net/browse/ATB-1203)
[ATB 1204](https://govukverify.atlassian.net/browse/ATB-1204)
### What changed
Ensure the intervention history contains pertinent information from the TICF_ACCOUNT_INTERVENTIONS event.

Each intervention history event will be stored as a pipe separated string in a list within the db in the following format. To make it more useful when returned through the API, it would be beneficial to transform and return each string as qualified json object in the history array.

for example:
`1609462861000|TICF_CRI|01|01|CMS|1234567|1234567`

if originating_component_id and originator_reference_id was not present, the history string would be:

`1609462861000|TICF_CRI|01|01|||1234567`

N.B. Create the '|' separator as a constant to ensure its consistent across lambdas

### Why did it change
Improve traceability.
Make the API response easier to read
This would keep the size of the db item to a minimal whilst maintaining the history of interventions place on the account.
## Testing
Locally, via unit tests:
All tests passing:
Screenshot 2023-11-27 at 15 59 30
Via API Gateway:
When passed from the ingress queue, correctly formats the history string:
Screenshot 2023-11-28 at 14 58 25
Screenshot 2023-11-28 at 14 58 50
Screenshot 2023-11-28 at 14 59 07
Transforms the string into an object:
Screenshot 2023-11-27 at 16 14 05
Screenshot 2023-11-27 at 15 48 35
Retains previous functionality of query parameters:
Screenshot 2023-11-27 at 15 48 53
Returns empty array if there is no history on the associated account:
Screenshot 2023-11-27 at 15 49 28

correctly filters out optional parameters in the history field if it is not present:
did not include originating_component_id and originator_reference_id in the history string:
Screenshot 2023-11-27 at 16 06 41
correctly filtered out:
Screenshot 2023-11-27 at 16 06 58

correctly applies logic to history field, able to return any valid history and filter out any invalid strings:
Screenshot 2023-11-27 at 16 35 21

Screenshot 2023-11-27 at 16 35 32

sends metric on any invalid strings that are processed:
Screenshot 2023-11-27 at 16 36 43

## Checklists


TaherKapasi (changes requested):

src/commons/history-string-builder.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Method to get the history object and validates if it contains the correct amount of components.
+   * @param historyString - string passed in with seperators ('|') included. This will be
+   * recieved from DynamoDB.
+   * @returns - The history string transformed into an object.
+   */
+  public getHistoryObject(historyString: string): HistoryObject {
+    const array = historyString.split(SEPARATOR);
+    const historyStringPartsValidate = Object.keys(HistoryStringParts).length / 2;
```
chiarapascucci-gds, line 58: hey @liamquayle  - we need to change this as per Taher's comment. Can you please create a const in the constants file, called something like `expectedHistoryStringLength` , then do the calculation there and use that imported constant in this file

src/commons/history-string-builder.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Method to get the history object and validates if it contains the correct amount of components.
+   * @param historyString - string passed in with seperators ('|') included. This will be
+   * recieved from DynamoDB.
+   * @returns - The history string transformed into an object.
+   */
+  public getHistoryObject(historyString: string): HistoryObject {
+    const array = historyString.split(SEPARATOR);
```
KG700, line 57:
```suggestion
    const historyArray = historyString.split(SEPARATOR);
```
Just to make it a bit clearer

src/commons/history-string-builder.ts
KG700, line 58: Just for my understanding - why do we need to divide this by 2?

src/commons/history-string-builder.ts
```diff
@@ -0,0 +1,93 @@
+import { HistoryStringParts, SEPARATOR } from '../data-types/constants';
```
KG700, line 1: Maybe SEPARATOR would fit better as a class attribute since it's specific role is specifically to do with the history

src/commons/history-string-builder.ts
chiarapascucci-gds: because calling Object.keys() on an Enum object returns all the key + all the values (because of how Enums are represented in js), so to get only the actual "keys" you divide by 2

src/services/dynamo-database-service.ts
```diff
@@ -145,6 +144,6 @@ export class DynamoDatabaseService {
       logAndPublishMetric(MetricNames.DB_QUERY_ERROR_TOO_MANY_ITEMS);
       throw new TooManyRecordsError(errorMessage);
```
KG700, line 145: This isn't part of the PR, and I might be missing something, but why are we throwing a 'too many records' error here?

KG700 (approved):

src/services/dynamo-database-service.ts
chiarapascucci-gds: Because we only ever expecting a given user to have one row in our DB, so if multiple records are returned there is something wrong with that user / our solution

jessica-li-01 (approved):

PR #101, opened 2023-11-29
ATB-1212 | Metrics to maintain count of accounts suspended and account blocked

chiarapascucci-gds:
## Proposed changes
This PR merges changes necessary to maintain 2 new metrics:
- ACCOUNTS_BLOCKED: maintains the count of total number of blocked accounts
- ACCOUNTS_SUSPENDED: maintains the count of total number of suspended accounts

it also merge small change to the EVENT_DELIVERY_LATENCY metric value and related alarm so that it uses the latency in ms not seconds 

### What changed
- made changes related to EVENT_DELIVERY_LATENCY: logging the value in milliseconds now and changed alarms thresholds to reflect this
- added new file with helper functions that log the relevant increments/decrements to the ACCOUNTS_BLOCKED/ACCOUNTS_SUSPENDED metrics given the specific change happening 
- refactored interface usage in account state retriever to facilitate reuse of interfaces
- added unit tests to maintain coverage 

### Why did it change
Introduce ability to track total number of accounts suspended/blocked without having to do a table scan

### Issue tracking

- [ATB-1212](https://govukverify.atlassian.net/browse/ATB-1212)

## Testing
Coverage maintained
Screenshot 2023-11-29 at 14 40 14
deployed changes to AWS account and tested that:
- block account of new user, then unblock it, blocked accounts count equals 0 ✅ 
- suspend account of new user, then unsuspend it, suspended accounts count equals 0 ✅ 
- suspend account with id and psw reset required, then send a psw reset successful event, suspend accounts count equals 1 ✅ 
- on that same user send a id reset successful event, suspend accounts count now equals 0 ✅ 
- suspend new user account, suspended accounts count is 1, and then block that same account, suspended account court goes to 0, blocked accounts count goes to 1 ✅ 

graphed view of relevant cloudwatch metrics (after several rounds of testing), as an example:

## Checklists


src/handlers/status-retriever-handler.ts
```diff
@@ -75,7 +75,7 @@ function validateEvent(userId: string) {
   return trimmedUserId;
 }
 
-function transformResponseFromDynamoDatabase(item: FullAccountInformation) {
+function transformResponseFromDynamoDatabase(item: Partial<FullAccountInformation>) {
```
liamquayle, line 78: Question: is this why the in the interface, the history has been changed to a required field, by utilising this Partial?

liamquayle (approved):

src/handlers/status-retriever-handler.ts
chiarapascucci-gds: Thank you for reviewing @liamquayle 🙏  
yes - essentially I needed an interface that would describe what we are expecting the get back from the DynamoDB when we get the full row for a given user. Given our implementation, if the user has a record in the DB, then we expect some fields to be present for sure (key field, the status flags, some timestamps etc). So I set those field as non-optional in the interface. But I then wrapped parameter type that this function take into a Partial, so that I get back a "copy" of that interface where all fields have been marked as optional, so that we can still use it the way it was being used here :)

TaherKapasi (approved):
nice ;-)

PR #152, opened 2024-01-09
ATB-1391 | Added methods for dynamo db manipulation

liamquayle:
## Proposed changes
[ATB-1391](https://govukverify.atlassian.net/browse/ATB-1391)

### What changed
Added a new file to export methods to enable easier manipulation of DynamoDB records

### Why did it change
Makes QA processes easier

## Testing
Locally:
Prints out full record when console.log is used:

Screenshot 2024-01-09 at 11 01 49

Update method:
Validated with Get call:

Screenshot 2024-01-15 at 14 53 53

Changes seen in DynamoDB:

Screenshot 2024-01-15 at 14 54 39

Correctly adds history as a list:
Screenshot 2024-01-15 at 15 12 03

Deleting a record from dynamoDB:

Screenshot 2024-01-09 at 11 11 30

Screenshot 2024-01-09 at 11 11 07

## Checklists

[ATB-1391]: https://govukverify.atlassian.net/browse/ATB-1391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
```diff
@@ -3,6 +3,7 @@ import { generateRandomTestUserId } from '../../../utils/generate-random-test-us
 import { sendSQSEvent } from '../../../utils/send-sqs-message';
 import { invokeGetAccountState } from '../../../utils/invoke-apigateway-lambda';
 import { timeDelayForTestEnvironment } from '../../../utils/utility';
+//import { updateAndRevert, getCurrentInformation, deleteTestRecord } from '../../../utils/dynamodb-update-revert';
```
ChrisBates1, line 6: Hi Liam, can remove this line if the imports are not being used please

ChrisBates1 (approved):

BhanusriNarne (approved):

feature-tests/utils/dynamodb-update-revert.ts
```diff
@@ -0,0 +1,109 @@
+import { DynamoDB, UpdateItemCommand, UpdateItemCommandInput } from '@aws-sdk/client-dynamodb';
+import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
```
chiarapascucci-gds, line 2: Hey Liam - I found this link that explains what I mentioned briefly in dev techy: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/introduction/
You are importing the dynamoDB library in two different way here:
- modular sdk v3 way where you import only the basic client and the individual commands you need
- you are also importing `DynamoDB` which I think is an class representation of the client that under the hood will import all possible operations that you could do on DynanoDB, resulting in more code being imported.
From my understanding this will have an impact on performance and how large the build for the feature tests is, but in the grand scheme of things this is probably alright, not a massive issue :) just wanted to flag it. I would personally stick to the modular set up like we have in our DynamoDB service class, but as you said you were you trying something different

liamquayle: this is a very good point, will probably change it to the dynamoDBClient rather than importing the other one to increase performance, didn't even consider that initially, thank you for adding this along with the link!

chiarapascucci-gds (dismissed):

feature-tests/utils/utility.ts
```diff
@@ -22,3 +24,24 @@ export interface CurrentTimeDescriptor {
...
+  reproveIdentity: boolean;
+  history: string;
+  auditLevel: string;
+  isAccountDeleted?: boolean;
+  reprovedIdentityAt?: number;
+  resetPasswordAt?: number;
+  accountDeletedAt?: number;
+};
+
+export interface UserInformationFromTable {
```
chiarapascucci-gds, line 45: hey Liam - I do not think this is right actually... Or at leat this is not the level that the interface should be at. This should describe a JSON object (like the one above) with the expected fields. The unmarshall call shouldn't return a Record type

ChrisBates1 (approved):

BhanusriNarne (approved):

feature-tests/utils/utility.ts
```diff
@@ -22,3 +22,38 @@ export interface CurrentTimeDescriptor {
...
+  pk: string;
+  updatedAt: number;
+  appliedAt: number;
+  sentAt: number;
+  intervention: string;
+  blocked: boolean;
+  suspended: boolean;
+  resetPassword: boolean;
+  reproveIdentity: boolean;
+  auditLevel: string;
```
chiarapascucci-gds, line 53: shouldn't Audit level be optional? Also, as an observation, isn't this interface very similar to the one above? Have you check if you can use one for both scenarios? :)

chiarapascucci-gds (approved):

feature-tests/utils/utility.ts
liamquayle: Is it optional now, must of misremembered that, good catch, Yeah they are, just sloppy from me i think aha ill go and fix this, thank you!

TaherKapasi (approved):

PR #198, opened 2024-01-29
ATB-1447 Remove history item if its age exceeds 2 years

amd-gds:
### What changed
1. Added env var to keep retention period
2. Because we're not allowed to do a SET and a REMOVE on the same attribute in one operation, it would have been inefficient time-wise to perform and wait for 2 dynamodb updates, hence the reason why I opted to compute the history list and replace it

### Why did it change
To adhere with the DI data retention period

### Issue tracking

- [ATB-1447](https://govukverify.atlassian.net/browse/ATB-1447)

## Testing
1. new env var present
Screenshot 2024-01-29 at 18 56 10

2. For testing purposes, set it to 86400 to make sure we have expired items
Screenshot 2024-01-29 at 18 57 14

3. All items in the history that will be removed:
Screenshot 2024-01-29 at 18 58 02

4. Event sent:
Screenshot 2024-01-29 at 19 04 57

5. Results:
Screenshot 2024-01-29 at 23 50 23

## Checklists
- [x] Checked SonarCube and ensured no code smells were added

[ATB-1447]: https://govukverify.atlassian.net/browse/ATB-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commons/build-partial-update-state-command.ts
+++ b/src/commons/build-partial-update-state-command.ts
@@ -3,6 +3,8 @@ import { UpdateItemCommandInput } from '@aws-sdk/client-dynamodb';
 import { AISInterventionTypes, EventsEnum, MetricNames } from '../data-types/constants';
 import { logAndPublishMetric } from './metrics';
 import { HistoryStringBuilder } from './history-string-builder';
+import { AppConfigService } from '../services/app-config-service';
+import { getCurrentTimestamp } from './get-current-timestamp';
 
 /**
  * Method to build a Partial of UpdateItemCommandInput
@@ -11,12 +13,14 @@ import { HistoryStringBuilder } from './history-string-builder';
  * @param currentTimestamp - timestamp of now in ms
  * @param interventionEvent - intervention type
  * @param interventionName - optional intervention name if the event was a fraud intervention
+ * @param historyList - list of history items
  */
 export const buildPartialUpdateAccountStateCommand = (
   finalState: StateDetails,
   eventName: EventsEnum,
   currentTimestamp: number,
   interventionEvent: TxMAIngressEvent,
+  historyList: string[],
   interventionName?: AISInterventionTypes,
 ): Partial<UpdateItemCommandInput> => {
   const eventTimestamp = interventionEvent.event_timestamp_ms ?? interventionEvent.timestamp * 1000;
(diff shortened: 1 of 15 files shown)


TaherKapasi (changes requested):

src/commons/build-partial-update-state-command.ts
```diff
@@ -65,18 +77,53 @@ export const buildPartialUpdateAccountStateCommand = (
...
+}
+
+/**
+ * Helper function to determine which history items have not exceeded the retention period.
+ * @param historyList - list of history items
+ */
+function extractValidHistoryItems(historyList: string[]) {
+  const historyStringBuilder = new HistoryStringBuilder();
+  const retentionPeriod = AppConfigService.getInstance().historyRetentionSeconds;
+  const currentTimeSeconds = getCurrentTimestamp().seconds;
```
TaherKapasi, line 120: could we not use currentTimestamp instead and then don't need to do import { getCurrentTimestamp } from './get-current-timestamp';

src/commons/build-partial-update-state-command.ts
```diff
@@ -65,18 +77,53 @@ export const buildPartialUpdateAccountStateCommand = (
...
+
+  return historyList.map((historyItem) => {
+    const historyObject = historyStringBuilder.getHistoryObject(historyItem);
+    const sendAtSeconds = Math.floor(new Date(historyObject.sentAt).getTime() / 1000);
+    if (sendAtSeconds + retentionPeriod >= currentTimeSeconds) {
```
TaherKapasi, line 125:
```suggestion
    if (sendAtSeconds + AppConfigService.getInstance().historyRetentionSeconds >= currentTimeSeconds) {
```

TaherKapasi (approved):

src/commons/build-partial-update-state-command.ts
```diff
@@ -65,18 +77,53 @@ export const buildPartialUpdateAccountStateCommand = (
...
+/**
+ * Helper function to determine which history items have not exceeded the retention period.
+ * @param historyList - list of history items
+ */
+function extractValidHistoryItems(historyList: string[]) {
+  const historyStringBuilder = new HistoryStringBuilder();
+  const retentionPeriod = AppConfigService.getInstance().historyRetentionSeconds;
+  const currentTimeSeconds = getCurrentTimestamp().seconds;
+
+  return historyList.map((historyItem) => {
```
chiarapascucci-gds, line 122: I may be wrong, but I this that map here will return an array of `{ S: string }` mixed with undefined (when the condition is not met), I think those are ignored downstream, but may be worth using a `reduce` instead so we filter + map at the same time, to avoid any unexpected behaviour?

liamquayle (approved):

chiarapascucci-gds (approved):

PR #224, opened 2024-02-21
ATB-1498: Refactored metric logging so to publish metric only once

chiarapascucci-gds:
## Proposed changes
This PR merges changes to refactor the way we flush in-code custom metrics, so that we do not make a call to cloudwatch after every time we add a metric

### What changed
- removed call to `publishStoredMetics()` from metric helper method and renamed method to flag that now it only adds a metric
- removed log line from inside the same method, as we do not need to log metric to splunk anymore
- added call to flush metric in lambda handlers so to publish metric as few times as possible
- modified unit tests to check that we call `publishStoredMetics()` once per execution on all paths
- expanded delete processor unit test to test happy path, and paths with multiple records 
- added logging of error for publishing to txma errors
- removed debug line from status retriever that was logging the entire apig event, as this will contain user ID - not sure if there is a reason why it would be needed?

### Why did it change
increase efficiency and avoid cloudwatch throttling

### Issue tracking
- [ATB-1498](https://govukverify.atlassian.net/browse/ATB-1498)

## Testing
Unit tests passing and coverage maintained
deployed changes to AWS 
intervention processor
image
status retriever
Screenshot 2024-02-21 at 14 26 55
deletion 
image

## Checklists
- [x] Checked SonarCube and ensured no code smells were added

[ATB-1498]: https://govukverify.atlassian.net/browse/ATB-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/tests/account-deletion-processor-handler.test.ts
```diff
@@ -4,13 +4,19 @@ import logger from '../../commons/logger';
 import 'aws-sdk-client-mock-jest';
 import type { SQSEvent, SQSRecord } from 'aws-lambda';
 import { ContextExamples } from '@aws-lambda-powertools/commons';
+import { Metrics } from "@aws-lambda-powertools/metrics";
+import {MetricNames} from "../../data-types/constants";
```
KG700, line 8:
```suggestion
import { MetricNames } from "../../data-types/constants";
```
😛

src/handlers/account-deletion-processor-handler.ts
```diff
@@ -73,7 +74,8 @@ async function updateDeleteStatusId(userId: string) {
     await ddbService.updateDeleteStatus(userId);
   } catch (error) {
     logger.error(`${LOGS_PREFIX_SENSITIVE_INFO} Error updating account ${userId}`, { error });
-    logAndPublishMetric(MetricNames.MARK_AS_DELETED_FAILED);
+    addMetric(MetricNames.MARK_AS_DELETED_FAILED);
+    metric.publishStoredMetrics();
```
TaherKapasi, line 78: wondering if we could do try {} catch (e) {} finally { metric.publishStoredMetrics(); } so as to not do metric.publishStoredMetrics(); in try and catch separately

TaherKapasi: just a suggestion ;-)

KG700 (approved):

amd-gds (approved):

src/handlers/account-deletion-processor-handler.ts
chiarapascucci-gds: we could.. but then this would mean that if we are processing multiple records successfully, we would publish once per record, as opposed to one unique call when Promise.all resolved... whereas this way we publish once when an error is thrown, which interrupts the flow (so the call after promise.All doesn't execute bc the error is thrown)

chiarapascucci-gds: hope this makes sense @TaherKapasi :)

liamquayle (approved):

PR #233, opened 2024-02-28
ATB 1515 | Investigate and implement using Husky for pre-push hooks

liamquayle:
## Proposed changes
[ATB-1515](https://govukverify.atlassian.net/browse/ATB-1515)

### What changed
Enabled Husky to lint and fix code on `git commit` and to run `yarn test` during `git push`. 

### Why did it change
Get feedback faster if changes have broken unit tests

### Tested locally:
Runs `pre-commit` package on commit and unit tests on `git push`:
Screenshot 2024-03-04 at 14 51 37

## Checklists

[ATB-1515]: https://govukverify.atlassian.net/browse/ATB-1515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-yarn lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,17 +2,3 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn lint-staged
-
-git diff --cached --name-only | if grep --quiet template.yaml
-then
-  echo "\nRunning Checkov hook...\n"
-  checkov -d . --framework cloudformation --quiet --skip-check CKV_AWS_116
-fi
-
-git diff --cached --name-only | if grep --quiet template.yaml
-then
-  echo "\nRunning sam validate --lint...\n"
-  sam validate --lint
-fi
-
-exit 1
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -63,5 +66,9 @@
   },
   "resolutions": {
     "axios": "^1.6.5"
+  },
+  "lint-staged": {
+    "*": "prettier --write",
+    "*.ts": "yarn lint:code --fix"
```
chiarapascucci-gds, line 72: can we check that this does not do lint fixes on the tests files? There are some stuff that linting changes in the tests that can break things

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
```
chiarapascucci-gds, line 3: I would not expect husky to be used for pre-commit, just pre-push, pre-commit we want to keep our current solution

package.json
liamquayle: it will actually lint the test files so good catch, found a workaround though so will add that in

chiarapascucci-gds: nice :) also make sure it picks up our lint configurations and not just lints from another set of rules

chiarapascucci-gds: or actually.. I would leave the linting in the pre-commit, if we have it in the pre-commit, then I do not see a reason why we should repeat it pre-push

chiarapascucci-gds (approved):

TaherKapasi (approved):

PR #257, opened 2024-03-14
ATB-1545 updated state machine with operations intervention

KG700:
## Proposed changes

[ATB-1545 | Spike | Investigate changes to state machine](https://govukverify.atlassian.net/browse/ATB-1545)

### What changed

- Updated state machine config with Operations edge, and the allowable interventions for this edge.
- Added updated state machine unit tests for the scenarios for the operations intervention
- Ensured existing feature tests are still passing

### Why did it change

To enable TSD to use the AIS to trigger a user's identity reset intervention.

## Checklists


amd-gds (approved):

src/services/account-states/config.ts
```diff
@@ -82,6 +82,11 @@ export const transitionConfiguration: TransitionConfigurationInterface = {
       name: EventsEnum.FRAUD_UNBLOCK_ACCOUNT,
       interventionName: AISInterventionTypes.AIS_ACCOUNT_UNBLOCKED,
     },
+    '25': {
```
chiarapascucci-gds, line 85: is there a reason for using 25? I would suggest continuing from 89

KG700: Obviously the '25' is up for debate, but our reasoning was:
- the '2' of the '25' was chosen so if any more fraud events are added, this provides a fair bit of space to allow fraud events to all be grouped together (there could be about 20 fraud events)
- the '5' of the '25' sort of links it to the '05' event (since they're both identity reset)

chiarapascucci-gds: yeah that makes sense, but the thing is these numbers do not really mean anything, I think there is a more realistic risk (however unlikely) that we may need to add interventions in the future, where fraud does make use of those numbers.. and if we ever get up to 25 intervention then we would need to stop fraud using that number, because we put something there. Does that make sense? The current approach is that anything that is not fraud related (where the numbers mean something)

KG700: So did fraud choose the codes '01' - '07'? If so then I see your point. What about, though, if the account interventions service starts being used for multiple services, if we start counting backwards from 89 the other services' codes won't be grouped nicely. How about if we chose a number that's higher, like '75' (and we could say the '7' represents the service operations team and the '5' represents the identity reset event), that should prevent any possible clashes happening with future fraud codes, but maintains the ability to group thing tidily in the future)

chiarapascucci-gds: yep, Fraud send us code 01 through 07 and each is mapped to a specific intervention. Anything else we put to the end (99 backwards) so to reduce the risk of overlapping with Fraud codes to the minimum. I do not really see how anything else apart from the fraud codes are grouped together? I would see it as "codes that come from the event producer" v anything else, thus why I think this new code should be grouped as anything else. Maintaining our own internal categorisation is just extra effort that does mean anything outside of our system.

KG700: The way I see it is the codes that start with '9' are resolution events, and the Fraud events are trigger events. And since the Tsd event is also a trigger event it should be separate from the '9' events.

KG700: But anyway, this conversation is very much about what the requirements should be and this PR is only part of a spike that's intended to check we understand what changes need to be made to the logic in order to implement this change. So I suggest we park this conversation and have this discussion with the rest of the team later.

PR #273, opened 2024-03-20
ATB-1569: changing error type for TRANSITION_SAME_AS_CURRENT_STATE

chiarapascucci-gds:
## Proposed changes
Changed error type of the error being raised when an intervention has been applied but the resulting state is the same as the original state, which is not an expected behaviour of the state engine. If this situation arises, a `StateEngineConfigurationError` will be thrown and message will be retried.
Additionally, this PR adds logic NOT to publish a STATE_TRANSITION_NOT_ALLOWED_OR_IGNORED metric when a user led action event is received on an account that has no active interventions on it

### What changed
- changed type of error being thrown for the situation described above
- add a new test file that test the structure of the config file being used, specifically it tests that:
        - there are no state nodes with the same `blocked` `suspended` `passwordReset` `identityReset` settings
        - node to adj list is a 1:1 relation
        - all edges point to an existing node
some of these checks are already performed inside the state engine code, but this new file provides a quick way of checking that any config changes comply with the structure the config needs to have for the engine to work
- added check to in the `buildStateTransitionError` method to see if the initial state of the account was okay (all flags were set to false) and if the intervention received is a user led action, if this is NOT the case then a metric is logged together with an error message

### Why did it change
reduce risk of breaking changes

### Issue tracking

- [ATB-1569](https://govukverify.atlassian.net/browse/ATB-1569)

## Testing
Deployed changes to dev account and run feature tests, all passing

sent a psw reset event and id reset event on a non-existing account
Screenshot 2024-03-21 at 14 34 53

Screenshot 2024-03-21 at 14 35 53

sent a psw reset event and id reset event on an existing account in a unsuspended state ✅ same behaviour

sent psw reset and id reset on an account in suspended state (as per current implementation)
Screenshot 2024-03-21 at 14 45 02
Screenshot 2024-03-21 at 14 46 20

## Checklists

[ATB-1569]: https://govukverify.atlassian.net/browse/ATB-1569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/test/account-state-engine.test.ts
```diff
@@ -1,4 +1,4 @@
-import { AccountStateEngine } from '../account-states/account-state-engine';
+import {AccountStateEngine, areAccountStatesTheSame, compareStrings} from '../account-states/account-state-engine';
```
TaherKapasi, line 1: spaces

TaherKapasi (approved):

TaherKapasi (approved):

amd-gds (approved):

TaherKapasi (approved):

PR #290, opened 2024-03-26
ATB 1544 Refactoring of send audit events

amd-gds:
## Proposed changes
Refactoring of the send audit events method. Raised this PR to check the approach

### What changed
Unit tests not working - looking into fixing

### Issue tracking

- [ATB-1544](https://govukverify.atlassian.net/browse/ATB-1544)

## Checklists

[ATB-1554]: https://govukverify.atlassian.net/browse/ATB-1554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ATB-1544]: https://govukverify.atlassian.net/browse/ATB-1544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/interventions-processor-handler.ts
```diff
@@ -135,12 +143,21 @@ async function handleError(error: unknown, record: SQSRecord) {
...
+    // await (recordBody.event_name === TriggerEventsEnum.OPERATIONAL_ACCOUNT_INTERVENTION
+    //   ? sendAuditEvent(
+    //       TxMAEgressEventName.AIS_NON_FRAUD_EVENT_TRANSITION_IGNORED,
+    //       error.transition,
+    //       recordBody,
+    //       error.output,
+    //     )
+    //   : sendAuditEvent(TxMAEgressEventName.AIS_EVENT_TRANSITION_IGNORED, error.transition, recordBody, error.output));
   } else {
     logger.error('Error caught, message will be retried.', { errorMessage: (error as Error).message });
```
KG700, line 162: If you update the type for the `error` parameter for this function to this:
```
async function handleError(error: ValidationError | TooManyRecordsError | StateTransitionError | Error, record: SQSRecord) {..}
```
Which makes sense more sense from a Typescript perspective since the error type isn't technically unknown to us. Then you don't need to cast the error type to Error for the `else` block.
```suggestion
    logger.error('Error caught, message will be retried.', { errorMessage: error.message });
```

src/handlers/interventions-processor-handler.ts
```diff
@@ -180,11 +200,24 @@ async function validateAccountIsNotDeleted(
   if (itemFromDB?.isAccountDeleted === true) {
```
KG700, line 200:
```suggestion
  if (itemFromDB?.isAccountDeleted) {
```

src/services/account-states/config.ts
```diff
@@ -118,6 +118,6 @@ export const transitionConfiguration: TransitionConfigurationInterface = {
     AccountIsSuspended: ['02', '03', '04', '05', '06'],
     AccountNeedsPasswordReset: ['01', '02', '03', '05', '06', '25', '90', '94'],
     AccountNeedsIdReset: ['01', '02', '03', '04', '06', '91'],
-    AccountNeedsPswAndIdReset: ['01', '02', '03', '04', '05', '92', '93', '95'],
+    AccountNeedsPswAndIdReset: ['01', '02', '03', '04', '05', '25', '92', '93', '95'],
```
KG700, line 121: I _think_ the '25' shouldn't be in the `AccountNeedsPswAndIdReset`? Or are we going to fix that with the tweaks I've done on my branch?

src/handlers/interventions-processor-handler.ts
TaherKapasi: Hi @KG700 , @amd-gds - leave as-is just to make it explicit otherwise it be misinterpreted AND to guard against any non-falsy values inadvertently being interpreted as true.

src/handlers/interventions-processor-handler.ts
```diff
@@ -116,7 +121,7 @@ async function processSQSRecord(record: SQSRecord) {
 
   updateAccountStateCountMetric(currentAccountState, statusResult.stateResult);
   addMetric(MetricNames.INTERVENTION_EVENT_APPLIED, [], 1, { eventName: eventName.toString() });
-  await sendAuditEvent('AIS_EVENT_TRANSITION_APPLIED', eventName, recordBody, statusResult);
+  await new AuditEvents(TxMAEgressEventTransitionType.TRANSITION_APPLIED, eventName, recordBody, statusResult).send();
```
chiarapascucci-gds, line 124: Should this class maybe be called SendAuditEventsService?

src/handlers/interventions-processor-handler.ts
```diff
@@ -135,12 +140,13 @@ async function handleError(error: unknown, record: SQSRecord) {
     });
   else if (error instanceof StateTransitionError) {
     logger.warn('StateTransitionError caught, message will not be retried.', { errorMessage: error.message });
-    await sendAuditEvent(
-      'AIS_EVENT_TRANSITION_IGNORED',
+    const recordBody = JSON.parse(record.body) as TxMAIngressEvent;
+    await new AuditEvents(
+      TxMAEgressEventTransitionType.TRANSITION_IGNORED,
```
chiarapascucci-gds, line 145: are we changing the event names from 'AIS_EVENT_TRANSITION_IGNORED' to just 'TRANSITION_IGNORED'?

src/handlers/interventions-processor-handler.ts
```diff
@@ -154,12 +160,15 @@ async function handleError(error: unknown, record: SQSRecord) {
...
+  if (
+    recordBody.event_name === TriggerEventsEnum.TICF_ACCOUNT_INTERVENTION ||
+    recordBody.event_name === TriggerEventsEnum.OPERATIONAL_ACCOUNT_INTERVENTION
+  ) {
     validateInterventionEvent(recordBody);
     const interventionCode = recordBody.extensions!.intervention!.intervention_code;
     return accountStateEngine.getInterventionEnumFromCode(interventionCode);
   }
-  return recordBody.event_name as unknown as EventsEnum;
+  return recordBody.event_name.toString() as EventsEnum;
```
chiarapascucci-gds, line 171: is there an advantage in doing this? event_name will already be a string, so is this operation neeeded?

src/handlers/interventions-processor-handler.ts
amd-gds: We're only changing it here, not on the message to send to txma.
Before sending it to txma it is manipulated according to the ingress event source (fraud/ non fraud)
see `extractEgressEventName` in AuditEvents

src/handlers/interventions-processor-handler.ts
amd-gds: we can either use .toString() or 
`recordBody.event_name as unknown as EventsEnum`
The compiler is complaining otherwise 
`Conversion of type TriggerEventsEnum to type EventsEnum may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to unknown first`

src/services/audit-events-service.ts
```diff
@@ -0,0 +1,197 @@
...
+        : 'AIS_EVENT_';
+
+    return (prefix + this.egressEventTransitionType) as TxMAEgressEventName;
+  }
+
+  /**
+   * Helper function to build extension object based on the type of event
+   * @returns - TxMAEgressExtensions object
+   */
+  private buildExtensions() {
```
chiarapascucci-gds, line 112: just out of curiosity, is there a reason for storing all of this info as class variables? :) I personally maybe find it a bit hard to read with all the `this` references, and I am also wondering if there may be some small efficiency gains in having this class as a singleton and just have those class variables as parameters to a method :)

src/handlers/interventions-processor-handler.ts
chiarapascucci-gds: yeah exactly, maybe toString() is easier to read then doing `unknown` but it is a method call that we can avoid by just doing that explicit type declaration for the compiler :)

amd-gds: I'll loop in @KG700 here for reference. I can't remember why we wanted to call `.toString()` instead of casting to `unknown`?

PR #308, opened 2024-04-26
ATB-1650 | Update API Spec and Lambda Permissions to enable Lambda Rollback through ClickOps

MinaMahdi:
## Proposed changes

ATB-1650 | Update Api Spec and Lambda Permissions to enable Lambda Rollback through ClickOps

### What changed

- Updated API Spec to reference to live instead of latest  
- Updated `AWS::Lambda::Permission` resource to reference Alias at the `FunctionName` property so that when the Trigger  is attached to our Alias version,  This is important so that if we ever need to rollback our Lambda / new version is created, the trigger will be attached to the correct version of the Lambda function and uses the latest version rather than any older version so no unexpected behaviour can take place. 

### Why did it change

Testing out ClickOps as a Lambda Rollback Option 

### Issue tracking

- [ATB-1650](https://govukverify.atlassian.net/browse/ATB-1650)

## Testing

**Status Retriever** 
Prior to Rollback:
Screenshot 2024-04-26 at 13 49 32
After changing Alias - Rollback : 
Screenshot 2024-04-26 at 13 50 32

**Interventions Processor**
Prior to Rollback: 
Screenshot 2024-04-26 at 13 40 21
After changing Alias - Rollback: 
Screenshot 2024-04-26 at 13 42 12

**Account Deletion** 

Prior to Rollback: 
Screenshot 2024-04-26 at 13 31 32

After changing Alias - Rollback: 
Screenshot 2024-04-26 at 13 32 46

## Checklists

[ATB-1650]: https://govukverify.atlassian.net/browse/ATB-1650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/infra/main/template.yaml
+++ b/src/infra/main/template.yaml
@@ -127,7 +127,7 @@ Globals:
     Architectures:
       - arm64
     Tracing: Active
-    AutoPublishAlias: latest
+    AutoPublishAlias: live
     CodeUri: ../../handlers
     KmsKeyArn: !GetAtt LambdaEnvironmentVariableEncryptionKey.Arn
     Environment:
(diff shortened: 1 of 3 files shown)


src/infra/main/template.yaml
```diff
@@ -1203,6 +1213,8 @@ Resources:
             Location: "../../specs/api.yaml"
       Description: "Private REST API that provides method to interact with the Account Interventions Services"
       StageName: !Ref ApiStageName
+      OpenApiVersion: 3.0.3
+      PropagateTags: true
```
KG700, line 1217: Just wondering what the `OpenApiVersion` and `PropagateTags` are for

MinaMahdi: Hey @KG700 
So SAM creates a stage name by default so by having in `OpenApiVersion`  it removes that extra Stage that we have in our PrivateRestApi, but this would only work when deploying the API from scratch, so sadly at the moment having it in, doesn't do much, it doesn't remove the stage as it needs to be added from the beginning, but think good to keep in nonetheless. 

For `PropagateTags` they way I understand it is that when its set to `True` any tags that are applied to the APIGW are also applied to the other underlying resources in the stack.

TaherKapasi (approved):

KG700 (approved):

PR #468, opened 2024-11-14
OLH-2289 Alarms for the Account Interventions Events Queue

saralk:
### What changed

- Create new alarms for the account interventions queue
- Create parameters to share the queue names between the main and alarm stack
- Updated a dev dependency as there was a security notice

### Why did it change

- So that we can monitor the status of account interventions, and be infomred when there is an issue

### Issue tracking

https://govukverify.atlassian.net/browse/OLH-2289

## Testing

Verified that the alarms are created on dev, and that they go into alarm

## Checklists
- [X] Did not commit any not-required changes to the `src/infra/**/samconfig.toml`
- [X] Tested changes and included test evidence in the PR, if appropriate
- [X] Included all required tags and other properties for any new resources in the SAM template
- [X] Ensured that any new resources in the SAM Template follow appropriate naming conventions
- [X] Ensured that naming of new resources is compatible with deploying multiple stacks with custom stack names during development
- [X] Ensured that no log lines include PII or other sensitive data
- [X] Ensured that all commits in this PR are signed
- [X] Ensured appropriate code coverage is maintained by unit tests
- [X] Checked SonarCube and ensured no code smells were added

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,6 @@
   "resolutions": {
     "axios": "^1.7.7",
     "cookie": "^0.7.0",
-    "jsonpath-plus": "^10.0.0"
+    "jsonpath-plus": "^10.0.7"
   }
 }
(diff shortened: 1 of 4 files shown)


saralk:
@latifcabinetoffice potentially, but this is a standard practice across all the lambads. I suppose you could end up in a situation where the messages are piling up faster than the lambda can process it, and you'd only trigger the too many messages alarm

peterfajemisincabinetoffice (approved):

PR #835, opened 2026-03-27
FPAD-7613: Add DisableAlarms parameter

edlouth:
## Proposed changes

Add parameter to the main AWS template to disable alarms.  This is done this way as the current parameter `AlertingSNSLowAlertNotificationTopicARN` is a special type `AWS::SSM::Parameter::Value<String>` which fetches the actual value from a store, and will fail if set to empty or none. 

### Issue tracking

- [FPAD-7613](https://govukverify.atlassian.net/browse/FPAD-7613)

[FPAD-7613]: https://govukverify.atlassian.net/browse/FPAD-7613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/infra/main/template.yaml
+++ b/src/infra/main/template.yaml
@@ -44,6 +44,11 @@ Parameters:
     Description: 'The ARN of the Alerting SNS Low Alert Notification topic'
     Type: AWS::SSM::Parameter::Value<String>
     Default: '/ais-infra-alerting/SNS/LowAlertNotificationTopic/ARN' #pragma: allowlist secret
+  DisableAlarms:
+    Type: String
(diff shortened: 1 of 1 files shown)


ede-gds (commented):
It makes me nervous having a big kill-switch for all notifications (which could get turned on in prod without us knowing and cause us to miss an incident). Maybe a better approach is to have a switch for sending all notifications to a different channel which is explicitly for spammy events but which we can audit after the fact.

PR #873, opened 2026-04-13
FPAD-6038: State transition error messages

edlouth:
## What

Include the current state from the database in state transition error messages.

## Why

Make it easier to debug.

## Notes

This change requires fetching an additional `intervention` field from the database.

### Issue tracking

- [FPAD-6038](https://govukverify.atlassian.net/browse/FPAD-6038)

[FPAD-6038]: https://govukverify.atlassian.net/browse/FPAD-6038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/account-states/account-state-engine.ts
+++ b/src/services/account-states/account-state-engine.ts
@@ -71,14 +71,18 @@ export class AccountStateEngine {
    * @param event - EventEnum representation of event received
    * @param initialState - optional state object representation the current state of the account, it defaults to account unsuspended if nothing is passed
    */
-  applyEventTransition(event: EventsEnum, initialState: StateDetails): AccountStateEngineOutput {
+  applyEventTransition(
+    event: EventsEnum,
+    initialState: StateDetails,
+    interventionName: string | undefined,
+  ): AccountStateEngineOutput {
     const allowedTransitions = this.determineNextAllowableInterventions(initialState);
-    const transition = this.getTransition(allowedTransitions, event, initialState);
+    const transition = this.getTransition(allowedTransitions, event, initialState, interventionName);
     const newStateObject = this.getNewStateObject(transition);
     if (areAccountStatesTheSame(newStateObject, initialState))
       throw buildConfigurationError(
         MetricNames.TRANSITION_SAME_AS_CURRENT_STATE,
-        'Computed new state is the same as the current state.',
+        `Computed new state is the same as the current state. Current state: ${interventionName ?? '(empty)'}; Event: ${event}`,
       );
     return {
       stateResult: newStateObject,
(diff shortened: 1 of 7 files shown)


src/services/account-states/account-state-engine.ts
```diff
@@ -71,14 +71,18 @@ export class AccountStateEngine {
    * @param event - EventEnum representation of event received
    * @param initialState - optional state object representation the current state of the account, it defaults to account unsuspended if nothing is passed
    */
-  applyEventTransition(event: EventsEnum, initialState: StateDetails): AccountStateEngineOutput {
+  applyEventTransition(
+    event: EventsEnum,
+    initialState: StateDetails,
+    itemFromDB: DynamoDBStateResult | undefined,
```
ede-gds: **Suggestion:** `itemFromDB` is smelly to me - I don't think `applyEventTransition` cares that the item is from the database, and it also doesn't tell us anything about what the item is.  

It doesn't look like we need to pass around the whole object so I think we can get away with just passing in `currentState` or something similar, which hopefully makes the function interface more intuitive at the same time

src/services/account-states/account-state-engine.ts
```diff
@@ -71,14 +71,18 @@ export class AccountStateEngine {
...
+  ): AccountStateEngineOutput {
     const allowedTransitions = this.determineNextAllowableInterventions(initialState);
-    const transition = this.getTransition(allowedTransitions, event, initialState);
+    const transition = this.getTransition(allowedTransitions, event, initialState, itemFromDB);
     const newStateObject = this.getNewStateObject(transition);
     if (areAccountStatesTheSame(newStateObject, initialState))
       throw buildConfigurationError(
         MetricNames.TRANSITION_SAME_AS_CURRENT_STATE,
-        'Computed new state is the same as the current state.',
+        `Computed new state is the same as the current state. Current state: ${itemFromDB?.intervention ?? '(empty)'}; Event: ${event}`,
```
ede-gds: **Thought:** Not for this PR but typescript errors are really really crap at structuring information. If this was a log we could just add a dictionary to it with `{ currentState: itemFromDB.intervention }` and have it searchable etc. Maybe we should think about a dedicated error primitive for bubbling this data up to the lambda error handling and have that log it out reliably.

ede-gds (commented):
Looks good, one minor thing

src/services/account-states/account-state-engine.ts
edlouth: As you say one for another PR, however this could be nice little refactor to work on.

src/services/account-states/account-state-engine.ts
edlouth: Fixed

ede-gds (approved):

PR #896, opened 2026-04-21
BAU: Update github actions to latest versions

edlouth:
## What

Update to latest versions of some GitHub actions.

## Why

GitHub is complaining that we are using actions which use node version 20.


ede-gds (approved):
lgtm

PR #911, opened 2026-04-28
FPAD-7873: Manual fixes to event catalogue schemas

edlouth:
## What

Update compile-schema to manually edit the schemas to apply missing features.

## Why

This way we can simplify our schema validator to use `oneOf`.
This fix should mean that all our events pass this validation and we can then remove the hand-rolled version.
We can then type our events automatically from the JSON schema.

## Notes

Ideally this would be tested in `dev` or similar to see that it actually works.

### Issue tracking

- [FPAD-7873](https://govukverify.atlassian.net/browse/FPAD-7873)

[FPAD-7873]: https://govukverify.atlassian.net/browse/FPAD-7873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commons/compile-schema.ts
+++ b/src/commons/compile-schema.ts
@@ -13,10 +13,35 @@ import {
 const ajv = new Ajv({ allErrors: true });
 const ajv2019 = new Ajv2019({ allErrors: true });
 
-ajv2019.addSchema(TICF_ACCOUNT_INTERVENTIONSchema);
-ajv2019.addSchema(AUTH_PASSWORD_RESET_SUCCESSFULSchema);
-ajv2019.addSchema(AUTH_PASSWORD_RESET_SUCCESSFUL_FOR_TEST_CLIENTSchema);
-ajv2019.addSchema(IPV_ACCOUNT_INTERVENTION_ENDSchema);
+interface Schema {
+  properties: Record<string, unknown>;
+  required: string[];
+}
+
+const addEventMetadataToSchema = (schema: Schema, eventName: string) => ({
+  ...schema,
+  required: [...schema.required, 'event_id'],
+  properties: {
+    ...schema.properties,
+    event_name: {
+      type: 'string',
+      const: eventName,
+    },
+    event_id: {
+      type: 'string',
+    },
+  },
+});
+
+ajv2019.addSchema(addEventMetadataToSchema(TICF_ACCOUNT_INTERVENTIONSchema, 'TICF_ACCOUNT_INTERVENTION'));
+ajv2019.addSchema(addEventMetadataToSchema(AUTH_PASSWORD_RESET_SUCCESSFULSchema, 'AUTH_PASSWORD_RESET_SUCCESSFUL'));
+ajv2019.addSchema(
+  addEventMetadataToSchema(
+    AUTH_PASSWORD_RESET_SUCCESSFUL_FOR_TEST_CLIENTSchema,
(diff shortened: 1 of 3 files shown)


src/commons/compile-schema.ts
```diff
@@ -13,10 +13,32 @@ import {
...
-ajv2019.addSchema(TICF_ACCOUNT_INTERVENTIONSchema);
-ajv2019.addSchema(AUTH_PASSWORD_RESET_SUCCESSFULSchema);
-ajv2019.addSchema(AUTH_PASSWORD_RESET_SUCCESSFUL_FOR_TEST_CLIENTSchema);
-ajv2019.addSchema(IPV_ACCOUNT_INTERVENTION_ENDSchema);
+interface Schema {
+  properties: Record<string, unknown>;
+  required: string[];
+}
+
+const fixSchema = (schema: Schema, eventName: string) => ({
```
ede-gds: **Suggestion:** Can we call this something more specific like `addEventMetadataToSchema`?

src/services/test/validate-event.test.ts
```diff
@@ -54,19 +55,60 @@ describe('event-validation', () => {
...
+    }).not.toThrow();
+    expect(() => {
+      validateInterventionEvent(TxMAEvent as TxMAEvent);
+    }).not.toThrow();
+    expect(addMetric).not.toHaveBeenCalledWith('INVALID_EVENT_RECEIVED_EVENT_CATALOGUE', undefined, undefined, {
+      EVENT_NAME: 'AUTH_PASSWORD_RESET_SUCCESSFUL',
+    });
+  });
+
+  it('should not return anything as event is valid - raise metric for invalid event', () => {
```
ede-gds: **Issue:** I find this test string very hard to parse - what is this actually testing for? It's out of scope for this PR but I think all the test strings in this file could use some love - entirely selfish on my part but if I break a test in a PR I want to understand what I've broken ASAP so I can get on with fixing it 😅

src/services/validate-event.ts
```diff
@@ -45,7 +45,10 @@ export function validateEventAgainstSchema(interventionRequest: TxMAIngressEvent
  * @param interventionRequest - the TxMA event
  */
 export function validateInterventionEvent(interventionRequest: TxMAIngressEvent): void {
-  if (Number.isNaN(Number.parseInt(interventionRequest.extensions?.intervention?.intervention_code ?? ''))) {
+  if (
+    interventionRequest.extensions &&
+    Number.isNaN(Number.parseInt(interventionRequest.extensions.intervention?.intervention_code ?? ''))
+  ) {
```
ede-gds: **Question:** What's the motivation here? Isn't this doing exactly the same thing as before the change?

ede-gds (commented):
Looks great, just questions about the detail

src/commons/compile-schema.ts
edlouth: Done

src/services/test/validate-event.test.ts
edlouth: I have fixed some names to be more descriptive and added a separate `describe` for `validateInterventionEvent`.

src/services/validate-event.ts
edlouth: The behaviour changes if `extensions` is `undefined` from throwing to passing.  However I have reverted this as it isn't being called in the code in the same way as the tests.

kirstyallen1 (approved):

PR #914, opened 2026-04-29
BAU-FT: Revert uuid upgrade

edlouth:


feature-tests/package.json
```diff
@@ -21,7 +21,7 @@
     "eslint": "^10.2.1",
     "jest": "^30.2.0",
     "jest-cucumber": "^4.5.0",
-    "jest-junit": "^16.0.0",
+    "jest-junit": "^17.0.0",
```
kirstyallen1, line 24: question: Is this version change intended/will it cause any potential issues?

edlouth: Yes this is intended, the newer version of `jest-junit` doesn't use the version of `uuid` with the security issue.

kirstyallen1 (approved):

PR #916, opened 2026-04-29
FPAD-7815: Feature tests to vitest v2

edlouth:
## What

Feature tests moved to Vitest

## Why

Uses latest ESM and will resolve the `uuid` security vulnerability

## Notes

This has been tested on `dev` however when it was last released it was a little flakey.

### Issue tracking

- [FPAD-7815](https://govukverify.atlassian.net/browse/FPAD-7815)

[FPAD-7815]: https://govukverify.atlassian.net/browse/FPAD-7815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


feature-tests/vitest.config.ts
```diff
@@ -0,0 +1,24 @@
...
+    include: ['**/*.step.ts'],
+    retry: 3,
+    hideSkippedTests: true,
+    testTimeout: 60_000,
+    env: {
+      TEST_ENVIRONMENT: 'dev',
+      SAM_STACK_NAME: 'ais-main',
+      AWS_REGION: 'eu-west-2',
+      AWS_PROFILE: 'dev',
+      tagFilter: '@regression',
```
ede-gds, line 20: **Suggestion:** I think if possible we should keep this configurable via environment variable (`process.env.tagFilter`)

ede-gds: I realised that maybe this gets overridden by environment variables anyway? I had a look at the vitest docs but couldn't work out if that's the case

feature-tests/tsconfig.json
```diff
@@ -13,7 +12,7 @@
     "transpileOnly": true,
     "files": true
   },
-  "include": ["tests", "apiEndpoints", "jest.config.ts", "utils", "jest-cucumber-config.js"],
+  "include": ["tests", "apiEndpoints", "utils"],
```
ede-gds, line 15: **Question:** Is `vitest.config.ts` included by these rules?

feature-tests/Dockerfile
```diff
@@ -1,27 +1,32 @@
...
 # Set up AWS CLI
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+RUN if [ "$TARGETARCH" = "arm64" ]; then \
+    URL="https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip"; \
+    else \
+    URL="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"; \
+    fi && \
+    curl -sL "$URL" -o "awscliv2.zip" && \
     unzip awscliv2.zip && \
     ./aws/install
```
ede-gds: **Suggestion:** Use `ARCH` so we can use `docker buildx` for multiarch stuff: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

**Issue:** Downloading something in a docker image and not validating it with a checksum is a nasty security hole - if someone takes control of `amazonaws.com` we're downloading and executing arbitrary code.

It might be better just to use the aws-cli image (although I can't work out what the base OS is for it): https://hub.docker.com/r/amazon/aws-cli

ede-gds (changes requested):

feature-tests/tsconfig.json
edlouth: Not currently, but I don't think it is needed.

feature-tests/Dockerfile
edlouth: `TARGETARCH` is the recommended way to do this and is compatible with `docker buildx` https://docs.docker.com/build/building/multi-platform/

edlouth: The aws-cli image is a great solution to this, it is based on the image we are currently using and gets rid of the multi-architecture issue entirely.

feature-tests/vitest.config.ts
edlouth: I believe they are overwritten, but I am also struggling to find any documentation.

feature-tests/utils/invoke-apigateway-lambda.ts
```diff
@@ -4,7 +4,7 @@ import request from 'supertest';
 import { App } from 'supertest/types';
 
 export async function invokeGetAccountState(testUserId: string, historyValue: boolean) {
-  if (process.platform === 'darwin') {
+  if (process.platform === 'darwin' || process.env['USE_PRIVATE_API_GATEWAY']) {
```
ede-gds, line 7: **Suggestion (non-blocking):** This isn't actually checking a platform-specific implementation, right? If I'm right and this is just a proxy for `are you a computer or a meatbag` I think we should just drop the platform check here, and have it behave as it should in AWS by default, with developers able to override it via environment variable.

ede-gds (approved):
lgtm, one minor thing

feature-tests/utils/invoke-apigateway-lambda.ts
edlouth: There are loads of `process.platform === 'linux'` snippets throughout the feature tests.  I think this is something for a new ticket to fix.

PR #936, opened 2026-05-14
FPAD-7970: Add design doc for persisting interventions

ede-gds:
## What

This is a design doc for persisting interventions in AIS.

### Issue tracking

- [FPAD-7970](https://govukverify.atlassian.net/browse/FPAD-7970)

[FPAD-7970]: https://govukverify.atlassian.net/browse/FPAD-7970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -27,6 +27,26 @@ Items are immutable once written.
 Two approaches are presented below.
 Both use DynamoDB and could coexist, but the choice of approach affects the table schema, the write logic, and how consumers query the data.
 
+#### General concerns
+
+##### Co-mingling data in the table
+
+Every case we discuss in this document involves using account ID as the partition key, and the current account status table also uses account ID as a partition key.
+This means a natural option is to store all rows in the same table and tag them so we know what the difference is.
+Consequences of this co-mingling approach compared with an approach using multiple tables are:
+
+- We don't need to create or maintain another table and the access to that table
+- We need to tag rows so we know what kind of object they represent
+- We need to handle existing rows which aren't tagged
+- We need to use the existing sort key on the account status table to do queries on interventions
+- We need to add logic to the application to handle the different kinds of rows and tags
+- We wouldn't be trivially able to throw all the intervention rows away and start again
+
+Based on these consequences, using a co-mingling approach would necessitate extra complexity in the application and the table itself while saving the cost of creating a new table.
+It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
+Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
+Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
+
 ---
 
 #### Option A: One mutable row per-intervention
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -39,6 +39,21 @@ Both use DynamoDB as per [programme guidelines](https://github.com/govuk-one-log
 The name of an intervention isn't recorded in the database.
 Where a name for an intervention or event is necessary based on this data, we will use a mapping recorded in code or configuration so we can easily update the name of an intervention.
 
+##### Versioning data
+
+DynamoDB doesn't enforce schemas so, aside from the partition key and the sort key, if we need to change the shape of the data we store in future we can do it entirely on the application side.
+For resilience and convenience, we should consider versioning the data, storing the version with the record itself.
+This version can be mapped to a parser at write time and read time, and used in the broader application for driving business logic.
+
+Versioning isn't necessary at this stage but should be implemented at the same time as the first change to the data schema.
+
+##### Ensuring limited records per-user
+
+To ensure the number of interventions needed for any given account stays small (and so the number of records queried for, sent over the wire, and stored and manipulated in memory), it may be necessary in future to truncate the table when querying.
+The easiest way to do this is to store a timestamp cursor for each account, for which we know we've dealt with all the interventions or events before.
+That way we can only retrieve records after the cursor, and can update the cursor using the intervention state after terminating a give intervention.
+The cursor design isn't part of this design - it has a performance impact and increases the complexity of the application, so should be implemented in future if there's a need for it.
+
 ---
 
 #### Option A: One mutable row per-intervention
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -21,6 +21,7 @@ This document proposes persisting interventions in a new DynamoDB table, written
 ## Assumptions
 
 - Interventions are independent between accounts
+- A given account has a small number of rows that we need to read at any one time. Small here means we don't have to worry about query performance and can store and manipulate them without having to worry about memory.
 
 ---
 
@@ -53,6 +54,13 @@ Based on these consequences, using a co-mingling approach would necessitate extr
 It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
 Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
 Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
+
+##### Interactions between interventions
+
+- superseding
+- sequencing
+- ownership
+- forbidden interventions
 
 ---
 
@@ -90,6 +98,10 @@ Secondary access patterns (e.g. reporting across all accounts) should be served 
 Events can arrive out of order or be replayed, in which case AIS might already have taken action before the "first" of two events arrives.
 This means for ordering purposes we should rely on when AIS receives an event, which most of the time will provide the same ordering as `sentAt`
 `sentAt` is preserved as a separate attribute for audit purposes.
+
+##### Consequences
+
+- We have more complicated queries (updates)
 
 ---
 
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -14,7 +14,7 @@ Without a more complete intervention history we cannot:
 - Report on intervention volumes, durations, and outcomes across the estate.
 - Support event-driven consumers — if AIS ever does need to act as a smart event bus, downstream services need a reliable ordered log of interventions to subscribe to.
 - Record if an intervention was removed due to it being `superseded`, `mitigated`, or `removed`.
-This document proposes persisting interventions immutably in new DynamoDB, written by the existing `InterventionsProcessorFunction` at the point a state transition is successfully applied.
+This document proposes persisting interventions immutably in a new DynamoDB table, written by the existing `InterventionsProcessorFunction` at the point a state transition is successfully applied.
 
 ---
 
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -21,7 +21,6 @@ This document proposes persisting interventions in a new DynamoDB table, written
 ## Assumptions
 
 - Interventions are independent between accounts
-- DynamoDB means we don't have to worry about scaling
 
 ---
 
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -1,5 +1,8 @@
 # Persisting Interventions
 
+Author: Ed Evans (ed.evans@digital.cabinet-office.gov.uk)
+Published: 2026-05
+
 ## Introduction
 
 AIS currently maintains account state in a single DynamoDB record per user (the `AccountStatusTable`).
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -37,24 +37,6 @@ Both use DynamoDB and could coexist, but the choice of approach affects the tabl
 
 #### General concerns
 
-##### Co-mingling data in the table
-
-Every case we discuss in this document involves using account ID as the partition key, and the current account status table also uses account ID as a partition key.
-This means a natural option is to store all rows in the same table and tag them so we know what the difference is.
-Consequences of this co-mingling approach compared with an approach using multiple tables are:
-
-- We don't need to create or maintain another table and the access to that table
-- We need to tag rows so we know what kind of object they represent
-- We need to handle existing rows which aren't tagged
-- We need to use the existing sort key on the account status table to do queries on interventions
-- We need to add logic to the application to handle the different kinds of rows and tags
-- We wouldn't be trivially able to throw all the intervention rows away and start again
-
-Based on these consequences, using a co-mingling approach would necessitate extra complexity in the application and the table itself while saving the cost of creating a new table.
-It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
-Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
-Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
-
 ##### Interactions between interventions
 
 - superseding
@@ -288,6 +270,27 @@ The relevant infrastructure will be deployed identically across environments mak
 ##### Mutation testing
 The new service is in scope for Stryker.
 The `persistIntervention` method and its callers should achieve the same mutation score threshold as the rest of the codebase.
+
+## Rejected Solutions
+
+### Co-mingling data in the table
+
+Every case we discuss in this document involves using account ID as the partition key, and the current account status table also uses account ID as a partition key.
+This means a natural option is to store all rows in the same table and tag them so we know what the difference is.
+Consequences of this co-mingling approach compared with an approach using multiple tables are:
+
+- We don't need to create or maintain another table and the access to that table
+- We need to tag rows so we know what kind of object they represent
+- We need to handle existing rows which aren't tagged
+- We need to use the existing sort key on the account status table to do queries on interventions
+- We need to add logic to the application to handle the different kinds of rows and tags
+- We wouldn't be trivially able to throw all the intervention rows away and start again
+
+Based on these consequences, using a co-mingling approach would necessitate extra complexity in the application and the table itself while saving the cost of creating a new table.
+It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
+Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
+Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
+
 
 ## Appendices
 
--- a/docs/design-docs/persisting-interventions.md
+++ b/docs/design-docs/persisting-interventions.md
@@ -1,0 +1,351 @@
+# Persisting Interventions
+
+## Introduction
+
+AIS currently maintains account state in a single DynamoDB record per user (the `AccountStatusTable`).
+That record is a mutable snapshot of account state: when a new intervention arrives, the previous state is overwritten.
+Each account has a JSON object storing the states it's moved through and some data about the relevant interventions, but no data about the states of the interventions themselvesn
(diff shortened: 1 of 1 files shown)


docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
+# Persisting Interventions
+
+## Introduction
+
+AIS currently maintains account state in a single DynamoDB record per user (the `AccountStatusTable`).
+That record is a mutable snapshot of account state: when a new intervention arrives, the previous state is overwritten.
+Each account has a JSON object storing the states it's moved through and some data about the relevant interventions, but no data about the states of the interventions themselvesn
+There is no durable log of what interventions were applied, when, or by whom.
```
edlouth: Is this true?  I think the existing `history` includes some of this information.

The current history is not kept forever, although this isn't different to this proposal.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+Without a more complete intervention history we cannot:
+
+- Support multiple interventions simultaneously on a single account without having `~O(n^2)` specific states for `n` interventions.
+- Add new interventions without adding `n` new states as above, and having a model for how that new intervention is applied.
+- Audit which interventions were applied to an account and in what order.
+- Report on intervention volumes, durations, and outcomes across the estate.
+- Support event-driven consumers — if AIS ever does need to act as a smart event bus, downstream services need a reliable ordered log of interventions to subscribe to.
+
```
edlouth, line 18:
```suggestion
- Record if an intervention was removed due to it being `superseded`, `mitigated`, or `removed`.
```

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+AIS currently maintains account state in a single DynamoDB record per user (the `AccountStatusTable`).
+That record is a mutable snapshot of account state: when a new intervention arrives, the previous state is overwritten.
+Each account has a JSON object storing the states it's moved through and some data about the relevant interventions, but no data about the states of the interventions themselvesn
+There is no durable log of what interventions were applied, when, or by whom.
+Without a more complete intervention history we cannot:
+
+- Support multiple interventions simultaneously on a single account without having `~O(n^2)` specific states for `n` interventions.
+- Add new interventions without adding `n` new states as above, and having a model for how that new intervention is applied.
+- Audit which interventions were applied to an account and in what order.
```
edlouth: I don't know if this is true

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+- Report on intervention volumes, durations, and outcomes across the estate.
+- Support event-driven consumers — if AIS ever does need to act as a smart event bus, downstream services need a reliable ordered log of interventions to subscribe to.
+
+This document proposes persisting interventions immutably in new DynamoDB, written by the existing `InterventionsProcessorFunction` at the point a state transition is successfully applied.
```
edlouth:
```suggestion
This document proposes persisting interventions immutably in a new DynamoDB table, written by the existing `InterventionsProcessorFunction` at the point a state transition is successfully applied.
```

edlouth: We could also:

1. Enrich the existing `history` field on the existing table.
2. Store a different type of row in the existing table (this is how FAI works).

edlouth: Does the new table need to be immutable?

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+| `interventionId`         | String        | —             | A generated UUIDv4. Provides a stable, globally unique identifier for this intervention record that can be referenced in audit trails, support tickets, and downstream systems without exposing the `accountId`. |
+| `interventionCode`       | String        | —             | The numeric code from `Codes` enum (e.g. `"01"`, `"02"`). Preserved verbatim from the ingress event. |
+| `interventionName`       | String        | —             | The resolved `AISInterventionTypes` enum value (e.g. `AIS_ACCOUNT_SUSPENDED`). Denormalised here so consumers do not need to join against config. |
+| `interventionReason`     | String        | —             | Free-text reason from `extensions.intervention.intervention_reason` in the ingress event. |
+| `interventionState`      | String        | —             | The current state of the intervention: `created`, `applied`, `superseded`, `mitigated`, or `removed`. |
```
edlouth, line 73: These don't match the statuses from [here](https://docs.google.com/presentation/d/1suP2v-YvJOY2q3X2957hAHIAzqkj6Z4lln5HhE9Isqo/edit?slide=id.g36f16a77f24_0_279#slide=id.g36f16a77f24_0_279)
Should we store `Ignored`?

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+| `originatingComponentId` | String        | —             | `extensions.intervention.originating_component_id`. Optional — not all events carry it. Test fixtures consistently show `component_id: 'TICF_CRI'` paired with `originating_component_id: 'CMS'`, confirming the two fields are not redundant. Both are preserved verbatim; the precise semantic distinction is owned by the upstream systems. |
+| `requesterId`            | String        | —             | `extensions.intervention.requester_id`. The identity of the human or system that requested the intervention - omitted for automated interventions. |
+| `originatorReferenceId`  | String        | —             | `extensions.intervention.originator_reference_id`. The reference ID in the originating system. |
+| `ttl`                    | Number        | —             | Unix timestamp in **seconds** at which DynamoDB should expire this item. Set to `createdAt / 1000 + retentionSeconds`. Retention period should match the programme's data retention policy (currently 2 years / 63,072,000 seconds, consistent with `DELETED_ACCOUNT_RETENTION_SECONDS`). |
```
edlouth: I think we want some sort of history or timestamps, specifically `removedAt`.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+
+##### Why `accountId` as partition key?
+The dominant access pattern is "give me all interventions for this account".
```
edlouth, line 82: The other main patterns are "give me all active interventions" and maybe "give me the highest priority intervention".

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+
+A dedicated KMS key (`InterventionStatusTableEncryptionKey`) with the same key policy pattern as `AccountStatusTableEncryptionKey` must be created alongside the table.
+The table ARN and encryption key ARN must be exported as CloudFormation outputs so the `ais-main` stack can import them.
+`InterventionsProcessorFunctionManagedPolicy` must be extended with `dynamodb:PutItem` on the new table ARN and `kms:Decrypt` / `kms:GenerateDataKey` on the new table's KMS key.
+No other Lambda roles require access to this table — it is write-only from the application's perspective.
+
+The option exists of using a dedicated queue and Lambda to decouple the write from the main processing path.
+This is probably unnecessary:
+
+- `InterventionsProcessorFunction` already performs a `dynamodb:UpdateItem` on `AccountStatusTable` synchronously in the hot path. Adding a second `dynamodb:PutItem` is the same class of operation with the same latency profile. It's difficult to find solid numbers for send latency on SQS but it's unlikely to be a critical difference.
```
edlouth, line 185: This is potentially more complicated for option A, as we may need to update existing interventions we will need to `Query` the existing interventions and then potentially `Put` or `Update` multiple interventions.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+- If the write fails, the existing retry mechanism (SQS `ReportBatchItemFailures`) will re-drive the message, so the history record will eventually be written when the state transition is retried.
+
+It may be worth considering using [DynamoDB transactions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transactions.html) to avoid the case where we only write successfully to either the account table or the interventions table.
```
edlouth, line 188: Depending on how complicated the new logic is, there may be an attraction to decoupling these, so that we aren't blocking `AccountStatusTable` updates if our new logic fails.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+Some quick googling implies this has performance impacts.
+An alternative is making the writes idempotent - this requires a read for every write so is probably not feasible.
+
+---
+
+### Interfaces
+
+We aim to have all persisted data strictly parsed at write time. 
+Because this data is entirely under our control (i.e. it's not part of the event catalogue and the schema isn't consumed by other components of DI) we can do this using Zod.
```
edlouth: Potentially, if we are strictly validating our inputs, it may not be necessary to revalidate before writing to the database.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+We should consider taking the opportunity to refactor how we communicate with DynamoDB to be re-usable and more in line with a ports-and-adapters approach, for increased flexibility and testability.
+
+Because of DynamoDB's lack of schema enforcement and the possibility of the data schema needing to evolve over time, we should consider versioning each row in the table and storing that version in the row itself.
+This allows us to parse based on version at write-time and read-time, and easily alter handling between versions and handle deprecation scenarios etc.
```
edlouth, line 201: At some point we should think a bit about how we are going to migrate data to this new world.  Presumably to start with we aren't going to assume any records exist in this new table, and drive entirely from the status in `AccountStatusTable`.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+##### No single point of failure
+DynamoDB is a managed, multi-AZ service.
+There is no additional availability concern introduced by this change.
+
+### Performance and Scalability
+
+##### Latency budget
+`InterventionsProcessorFunction` has a 5-second timeout.
+The existing processing path includes one `Query` and one `UpdateItem` against `AccountStatusTable`.
+Adding one `PutItem` against the history table adds approximately 2–5 ms at p99 (same-region DynamoDB).
```
edlouth, line 261: For Option A there are more operations required.

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,351 @@
...
+  originatorReferenceId?: string;
+  ttl: number;                 // seconds epoch
+}
+```
+
+##### Service interface
+
+```typescript
+export interface InterventionStatusService {
+  persistIntervention(record: InterventionHistoryRecord): Promise<void>;
```
edlouth, line 384: Will the logic inside this work out if an existing intervention exists and update or create accordingly?

edlouth, line 384: How will this work when `superseding` an intervention?  Will lots of the fields be ignored, or will those fields be overwritten?

docs/design-docs/persisting-interventions.md
ede-gds: this is definitely not true, and I even half rewrote this section to fix this already 🤦

docs/design-docs/persisting-interventions.md
ede-gds: yeah this is probably not true

docs/design-docs/persisting-interventions.md
ede-gds: nope, very much the opposite. Will fix

docs/design-docs/persisting-interventions.md
ede-gds: hm, ignored is an interesting case. I was using the simplified flow chart from the DRB check and forgot about it. I don't think we can answer the question about ignored until the ownership question is sorted, but I suppose in either case we probably store it for reporting reasons:

- if we politely ask other services to apply interventions, then it starts as `created` then as the first step moves to `applied` or `ignored` 
- if we rudely tell other services to apply interventions, then we just write `applied` or `ignored` straight up.

Adding the `applied` status in between `created` and the end states is I think becoming necessary, although I don't have an ironclad argument for it. One to discuss with Chris.

docs/design-docs/persisting-interventions.md
ede-gds: Can you be more specific? do you mean when the row gets removed for TTL reasons?

docs/design-docs/persisting-interventions.md
ede-gds: do you mean all active interventions for a given account? I don't think `highest priority intervention` is something dynamo can provide as part of this design.

docs/design-docs/persisting-interventions.md
ede-gds: yes, this is true

ede-gds: this is a hot mess actually - we need to grab all interventions history in case there's something we need to supersede right?

docs/design-docs/persisting-interventions.md
ede-gds: I think that's a big question - if our new table isn't a first-class source of truth, it's pretty awkward to be able to trust it to do anything. but if it is, we either need transactions or we need to cascade the operations in a sensible way

docs/design-docs/persisting-interventions.md
ede-gds: We might be talking at cross-purposes but the storage service doesn't know anything about inputs, aside from its own interface. Having a data-drive parsing layer before the write is in some ways about convenience - it gives us confidence we're getting the same things in and out, and that confidence is derived from data (Zod objects) instead of from code (the storage service's code)

docs/design-docs/persisting-interventions.md
ede-gds: Yeah, 100%. For now the only thing in scope is storing the things - at some point we'll have to move to a hybrid model where we check both tables, and hopefully eventually move to a model where we can trust just the intervention storage. I haven't thought through whether the account table is still worth having in that state

docs/design-docs/persisting-interventions.md
ede-gds: yeah, this needs much more thought

docs/design-docs/persisting-interventions.md
ede-gds: Probably not worth answering that question - this needs a lot of expanding 😅

ede-gds: see above 😅

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+- Report on intervention volumes, durations, and outcomes across the estate.
+- Support event-driven consumers — if AIS ever does need to act as a smart event bus, downstream services need a reliable ordered log of interventions to subscribe to.
+- Record if an intervention was removed due to it being `superseded`, `mitigated`, or `removed`.
+
+This document proposes persisting interventions in a new DynamoDB table, written by the existing `InterventionsProcessorFunction` at the point a state transition is successfully applied.
+
+## Assumptions
+
+- Interventions are independent between accounts
+- DynamoDB means we don't have to worry about scaling
```
edlouth: I think this is a given and can be removed

```suggestion
```

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+
```
edlouth, line 25:
```suggestion
- The number of interventions per account is not so large that we can't fetch all the interventions and handle them in code.
```

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+- We need to tag rows so we know what kind of object they represent
+- We need to handle existing rows which aren't tagged
+- We need to use the existing sort key on the account status table to do queries on interventions
+- We need to add logic to the application to handle the different kinds of rows and tags
+- We wouldn't be trivially able to throw all the intervention rows away and start again
+
+Based on these consequences, using a co-mingling approach would necessitate extra complexity in the application and the table itself while saving the cost of creating a new table.
+It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
+Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
+Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
```
edlouth: I think we are agreed on this, can we move it to the end as another option considered and rejected?

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+`sentAt` is controlled by the upstream system and is not guaranteed to be monotonically increasing.
+Events can arrive out of order or be replayed, in which case AIS might already have taken action before the "first" of two events arrives.
+This means for ordering purposes we should rely on when AIS receives an event, which most of the time will provide the same ordering as `sentAt`
+`sentAt` is preserved as a separate attribute for audit purposes.
+
+---
+
+#### Option B: Multiple immutable events per-intervention (event sourcing)
+
+Rather than storing one row per intervention, this approach stores one row per event that happens to an intervention.
```
edlouth, line 101: I presume this means that one "transaction" can produce multiple rows, for example a permanent suspension might undo a temporary suspension?

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+
+An intervention moves through a state machine:
+
+```
+created → applied → superseded
+                  → mitigated
+                  → removed
+```
+
+`applied` is a distinct state from `created` because application of an intervention may be asynchronous — the intervention can be recorded as `created` before AIS has confirmed it has taken effect.
```
edlouth, line 113: I think we would want to store a row for `ignored` events as well

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,378 @@
...
+PITR provides a continuous backup with a 35-day recovery window at no additional operational overhead.
+This is already enabled on `AccountStatusTable` and must be consistent here.
+
+##### DeletionProtection
+Must be enabled in all non-dev environments, consistent with `AccountStatusTable`.
+
+##### TTL-based expiry
+Items expire automatically via DynamoDB TTL.
+This is a soft delete — DynamoDB removes expired items asynchronously within 48 hours.
+The application must not rely on TTL for hard deletion guarantees; if regulatory requirements demand hard deletion within a specific window, a separate cleanup process is needed.
```
edlouth: At some point we need to specify how this works, I think for now "the same as the current `history`" will suffice

ede-gds:
wow i've definitely done something here

docs/design-docs/persisting-interventions.md
```diff
@@ -0,0 +1,417 @@
...
+- We need to tag rows so we know what kind of object they represent
+- We need to handle existing rows which aren't tagged
+- We need to use the existing sort key on the account status table to do queries on interventions
+- We need to add logic to the application to handle the different kinds of rows and tags
+- We wouldn't be trivially able to throw all the intervention rows away and start again
+
+Based on these consequences, using a co-mingling approach would necessitate extra complexity in the application and the table itself while saving the cost of creating a new table.
+It also removes our ability to trivially throw away the intervention rows while the table is write-only, which may be a useful step while iterating on the table specification.
+Setting up a new DynamoDB table is cheap and fairly simple, and we have existing infrastructure-as-code that we can use by analogy.
+Because of this, we're choosing a multiple table approach for this problem, and **not** using a co-mingling approach.
```
ParkRanger-Pro, line 317: (a reminder about the outstanding edit to this section)

```suggestion
Every option described here uses the account ID as the partition key. Currently the account status table also uses the same partition key. This means we could store the proposed data in the existing table. 

Whilst avoiding initial overhead in creating a new table, we prefer to create a new table for this data. This preference is because co-mingling would:

- introduce prefixing values of key(s), managing legacy records, and adding complex application logic to distinguish record types – work which would have little immediate value 
- inhibit throwing away the new table and start again, if the need occurred – something which may be valuable during early iteration of the schema design whist treat the data as write only

The approach now for using a new discrete table does not preclude co-mining data in the future. 
```

ede-gds:
@edlouth @ParkRanger-Pro This has become a bit of a millstone and isn't really helping so I'm just going to merge it. I'll write a separate doc about design doc semantics with something about how they're not living docs so we know not to rely on them past the implementation stage.

edlouth (approved):

PR #944, opened 2026-05-20
adr-003 zod validation

edlouth:
## What

Add an ADR for Zod validation

## Why

This seems like something to put in an ADR.

[FPAD-8081](https://govukverify.atlassian.net/browse/FPAD-8081)

[FPAD-8081]: https://govukverify.atlassian.net/browse/FPAD-8081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/adr/003-zod-validation.md
+++ b/adr/003-zod-validation.md
@@ -25,7 +25,7 @@ By utilising our own `Zod` types we can be sure that data enterring our system f
 
 ## Consequences
 
-- We will have control over our validation, independant of the event catalogue.
+- We will have control over our validation, independent of the event catalogue.
 - We won't be rejecting large numbers of processable messages.
 - We will get guaranteed type safety for all the data entering our system, based on automatically generated `Zod` types.
 - We have to write our own `Zod` validation in addition to the event catalogue schemas.
--- a/adr/003-zod-validation.md
+++ b/adr/003-zod-validation.md
@@ -1,0 +1,36 @@
+# Zod Validation
+
+## What
+
+Validate incoming messages from TxMA using the [Zod](https://zod.dev/) validation framework.
+Where messages fail validation we will block further processing and add a metric, so we can track the occurance.
+
+Limit validation to the properties that we require in our handlers.
+
+We will add non-blocking validation based on the JSON schemas from the event catalogue.
+When a message fails this validation we won't block further processing, but we will add a separate metric.
+
+We will aim to keep the `Zod` validation as a subset of the event catalogue JSON schemas.
+For now this will be manual and through unit testing with illustrative events, however long term we may aim to use a static tool.
+
+## Why
+
+It doesn't make sense to try and process messages which lack the required properties.
+
+Our preference would be to validate purely using event catalogue JSON schemas, which are a governed and shared resource.
+There is currently very limited upstream validation of the messages that we receive from TxMA.
+Our testing shows that many of the events we receive do not meet the event catalogue definitions.
+This means that we have low confidence that the messages we receive will always pass validation against the JSON schemas in the event catalogue.
+By utilising our own `Zod` types we can be sure that data enterring our system fits a known shape and is processable.
+
+## Consequences
+
+- We will have control over our validation, independant of the event catalogue.
+- We won't be rejecting large numbers of processable messages.
+- We will get guaranteed type safety for all the data entering our system, based on automatically generated `Zod` types.
+- We have to write our own `Zod` validation in addition to the event catalogue schemas.
+- By processing events that don't confirm to the event catalogue schemas, we may reduce how quickly components are fixed to send the correct events.
+
+## Alternatives
+
+Switch to utilising event catalogue schemas, however this would lead to large numbers of processable events being rejected.


adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
+# Zod Validation
+
+## What
+
+Validate incoming messages from TxMA using the [Zod](https://zod.dev/) validation framework. Where messages fail validation we will block further processing and add a metric, so we can track the occurance.
```
ede-gds: **Thought:** I need to write this down somewhere but when you're writing markdown can you split each sentence on to a new line? not so much for this PR but in general it can be a pain to review otherwise

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
...
+
+We will add non-block validation based on the JSON schemas from the event catalogue. When a message fails this validation we won't block further processing, but we will add a separate metric.
+
+## Why
+
+It doesn't make sense to try and process messages which lack the required properties.
+
+There is currently very limited upstream validation of the messages that we receive from TxMA. This means that we have low confidence that the messages we receive will always pass validation against the JSON schemas in the event catalogue.
+
+## Consequences
```
ede-gds, line 26: **Thought:** In the template these are very intentionally **not** split into good and bad, because (in my experience) it tends to lead to more nuance and a more complete list if you just write down the consequences. Not sure I'm going to insist on it here but maybe I'll update the template.

SImilarly, this section doesn't need to be an argument, because that's what the `why` section is for. Instead it's for encouraging thinking through of the value judgement-free broad consequences.

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
+# Zod Validation
+
+## What
+
+Validate incoming messages from TxMA using the [Zod](https://zod.dev/) validation framework. Where messages fail validation we will block further processing and add a metric, so we can track the occurance.
+
+Limit validation to the properties that we require in our handlers.
```
ede-gds, line 8: **Suggestion:** Can we add something in here about making an effort to ensure the Zod validation is a subset of the event schema? (what does that mean? maybe any event that passes the event catalogue layer should pass the Zod layer?)

Possible tools:

- static analysis
- property-based testing
- unit testing with illustrative events
- using our eyeballs and brains and writing comments

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
...
+
+We will add non-block validation based on the JSON schemas from the event catalogue. When a message fails this validation we won't block further processing, but we will add a separate metric.
+
+## Why
+
+It doesn't make sense to try and process messages which lack the required properties.
+
+There is currently very limited upstream validation of the messages that we receive from TxMA. This means that we have low confidence that the messages we receive will always pass validation against the JSON schemas in the event catalogue.
```
ede-gds: **Suggestion:** I think there's context missing here - something like 'we wish we could trust the event schemas because having a shared idea of what data is flying around one login would be great, but we can't so we're going to do something practical but also make sure we have a good idea of where both our parsing and the incoming messages diverge from the event catalogue's standards'

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
...
+
+- We will have control over our validation, independant of the event catalogue.
+- We won't be rejecting large numbers of processable messages.
+- We will get automatic Typescript types from `Zod`.
+
+### Negative
+
+- There is a risk that the `Zod` validation doesn't match or diverges from the event catalogue schemas.
+- We have to write our own `Zod` validation.
+- We may not be driving good behaviors in other teams.
```
ede-gds: **nitpick:**
```suggestion
- We may not be driving good behaviours in other teams.
```

**Thought:** I think we can frame this more concretely - something like 'Events we process which contravene the event catalogue schemas won't cause problems for the teams which emit them'

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Positive
+
+- We will have control over our validation, independant of the event catalogue.
+- We won't be rejecting large numbers of processable messages.
+- We will get automatic Typescript types from `Zod`.
+
+### Negative
+
+- There is a risk that the `Zod` validation doesn't match or diverges from the event catalogue schemas.
```
ede-gds: **Suggestion:** I'm not sure this this is a consequence - I think the consequence is 'We will process events which don't match the event schema', which might be covered by the `what` section. There might be a downstream consequence of the two diverging although I can't really articulate it?

adr/003-zod-validation.md
```diff
@@ -0,0 +1,33 @@
...
+
+There is currently very limited upstream validation of the messages that we receive from TxMA. This means that we have low confidence that the messages we receive will always pass validation against the JSON schemas in the event catalogue.
+
+## Consequences
+
+### Positive
+
+- We will have control over our validation, independant of the event catalogue.
+- We won't be rejecting large numbers of processable messages.
+- We will get automatic Typescript types from `Zod`.
```
ede-gds: **Suggestion:** I think this is underselling - what we actually get is guaranteed type safety for all the data entering our system via events which is pretty great.

ede-gds (commented):
Thanks for putting this in an ADR (genuinely great idea) and sorry it took me a couple of days to get to it. 

also sorry for the heap of comments - I do think this is a really good thing for us to do and clearer thinking at this point is probably a good thing so that's why I'm insisting on precision / being an ass

adr/003-zod-validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+Our preference would be to validate purely using event catalogue JSON schemas, which are a governed and shared resource.
+There is currently very limited upstream validation of the messages that we receive from TxMA.
+Our testing shows that many of the events we receive do not meet the event catalogue definitions.
+This means that we have low confidence that the messages we receive will always pass validation against the JSON schemas in the event catalogue.
+By utilising our own `Zod` types we can be sure that data enterring our system fits a known shape and is processable.
+
+## Consequences
+
+- We will have control over our validation, independant of the event catalogue.
```
ede-gds: **Nitpick:**
```suggestion
- We will have control over our validation, independent of the event catalogue.
```

ede-gds (approved):
lgtm, thanks!

PR #966, opened 2026-06-02
FPAD-8138: Typed table service

edlouth:
## What

A new database table class which validates returned rows against a `Zod` schema and gives correctly typed output.

## Why

DynamoDB doesn't have a fixed schema, so we need to handle scenarios where the data is not in the expected format.

## Notes

I have left the `AccountStatus` table as it is for now. 

I have deployed this to the `dev` environment and tested that this logs out correctly.

### Issue tracking

- [FPAD-8138](https://govukverify.atlassian.net/browse/FPAD-8138)

[FPAD-8138]: https://govukverify.atlassian.net/browse/FPAD-8138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/tests/interventions-processor-handler.test.ts
+++ b/src/handlers/tests/interventions-processor-handler.test.ts
@@ -12,6 +12,7 @@ import { AISInterventionTypes, EventsEnum, MetricNames, TriggerEventsEnum } from
 import { sendAuditEvent } from '../../services/send-audit-events';
 import { publishTimeToResolveMetrics } from '../../commons/metrics-helper';
 import { InterventionEventMessage, TicfAccountIntervention } from '../../contracts/intervention-events';
+import { InMemoryInterventionEventsService } from '../../tables/intervention-events';
 
 vi.mock('@aws-lambda-powertools/logger');
 vi.mock('../../commons/metrics');
@@ -92,6 +93,9 @@ const mockValidateEventAgainstSchema = vi.spyOn(validationModule, 'validateEvent
 
 const accountStateEngine = AccountStateEngine.getInstance();
 accountStateEngine.getInterventionEnumFromCode = vi.fn().mockImplementation(() => EventsEnum.FRAUD_BLOCK_ACCOUNT);
+
+const emptyInterventionEventsService = new InMemoryInterventionEventsService([]);
+
 describe('intervention processor handler', () => {
   let mockEvent: SQSEvent;
   let mockRecord: SQSRecord;
@@ -151,7 +155,7 @@ describe('intervention processor handler', () => {
   describe('handle', () => {
     it('does nothing if SQS event contains no record', async () => {
       const loggerWarnSpy = vi.spyOn(logger, 'warn');
-      await handler({ Records: [] }, mockContext);
+      await handler({ Records: [] }, mockContext, emptyInterventionEventsService);
       expect(loggerWarnSpy).toHaveBeenCalledWith('Received no records.');
       expect(addMetric).toHaveBeenCalledWith('INVALID_EVENT_RECEIVED');
       expect(publishTimeToResolveMetrics).not.toHaveBeenCalled();
@@ -159,7 +163,7 @@ describe('intervention processor handler', () => {
 
     it('should not retry if message body cannot be parsed to valid JSON', async () => {
       mockRecord.body = ' ';
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       // eslint-disable-next-line @typescript-eslint/unbound-method
@@ -188,7 +192,7 @@ describe('intervention processor handler', () => {
           interventionName: AISInterventionTypes.AIS_NO_INTERVENTION,
         });
       });
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       // eslint-disable-next-line @typescript-eslint/unbound-method
@@ -224,7 +228,7 @@ describe('intervention processor handler', () => {
         interventionName: EventsEnum.FRAUD_BLOCK_ACCOUNT,
         nextAllowableInterventions: [],
       });
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       expect(sendAuditEvent).toHaveBeenLastCalledWith(
@@ -261,7 +265,7 @@ describe('intervention processor handler', () => {
         interventionName: EventsEnum.FRAUD_BLOCK_ACCOUNT,
         nextAllowableInterventions: [],
       });
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       expect(sendAuditEvent).toHaveBeenLastCalledWith(
@@ -299,7 +303,7 @@ describe('intervention processor handler', () => {
       });
       mockRecord.body = JSON.stringify(resetPasswordEventBody);
       mockEvent.Records = [mockRecord];
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       expect(sendAuditEvent).toHaveBeenLastCalledWith(
@@ -342,7 +346,7 @@ describe('intervention processor handler', () => {
         suspended: false,
         isAccountDeleted: true,
       });
-      expect(await handler(mockEvent, mockContext)).toEqual({
+      expect(await handler(mockEvent, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [],
       });
       expect(addMetric).toHaveBeenLastCalledWith(MetricNames.ACCOUNT_IS_MARKED_AS_DELETED);
@@ -374,7 +378,7 @@ describe('intervention processor handler', () => {
       mockValidateEventAgainstSchema.mockReturnValueOnce(interventionEventBodyInTheFuture);
 
       mockRecord.body = JSON.stringify(interventionEventBodyInTheFuture);
-      expect(await handler({ Records: [mockRecord] }, mockContext)).toEqual({
+      expect(await handler({ Records: [mockRecord] }, mockContext, emptyInterventionEventsService)).toEqual({
         batchItemFailures: [
           {
             itemIdentifier: '123',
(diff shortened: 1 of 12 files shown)


src/tables/intervention-events.ts
```diff
@@ -0,0 +1,46 @@
...
+  interventionId: z.string(),
+  createdAt: z.number(),
+  interventionState: z.enum(InterventionState),
+  interventionCode: z.enum(Codes),
+  interventionReason: z.string(),
+  sentAt: z.number(),
+  componentId: z.string(),
+  originatingComponentId: z.string().optional(),
+  requesterId: z.string().optional(),
+  originatorReferenceId: z.string().optional(),
```
edlouth: `originatorReferenceId` is the field that TICF Ops are sending as `string[]` currently

edlouth: I have changed this to `z.union([z.string(), z.array(z.string())]).optional()` as otherwise later changes won't work.

src/services/db-table-service.ts
```diff
@@ -0,0 +1,37 @@
+import { z, ZodArray, ZodObject, ZodRawShape } from 'zod';
+import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
+import { BatchWriteCommand, DynamoDBDocumentClient, QueryCommand, QueryCommandInput } from '@aws-sdk/lib-dynamodb';
+
+type ArrayFromT<T extends ZodObject<ZodRawShape>> = z.infer<z.ZodArray<T>>;
+
+export abstract class DbTableService<T extends ZodObject<ZodRawShape> = ZodObject<ZodRawShape>> {
```
ede-gds: **Suggestion:** We should favour composition over inheritance wherever possible so this shouldn't be an abstract class, and instead it should implement an interface.

I think the naming should be something like 

```
export interface RecordService ...

export class DynamoDBRecordService implements RecordService ...
```

Then we can have a fake called something like `InMemoryRecordService` for using in tests

src/services/db-table-service.ts
```diff
@@ -0,0 +1,37 @@
...
+  readonly arraySchema: ZodArray<T>;
+  readonly docClient: DynamoDBDocumentClient;
+
+  public constructor(
+    readonly tableName: string,
+    readonly dbClient: DynamoDBClient,
+    readonly schema: T,
+  ) {
+    this.docClient = DynamoDBDocumentClient.from(dbClient);
```
ede-gds: **Suggestion:** I think if we need a `DynamoDBDocumentClient` we should just pass it in - if we want something more convenient we can always add a factory function.

src/handlers/tests/interventions-processor-handler.test.ts
```diff
@@ -19,6 +19,7 @@ vi.mock('@aws-sdk/util-dynamodb');
 vi.mock('../../services/dynamo-database-service');
 vi.mock('../../services/send-audit-events');
 vi.mock('../../commons/metrics-helper');
+vi.mock('../../tables/intervention-events');
```
ede-gds: **Suggestion:** I want to get away from runtime patching in tests - one way we can do this is changing the handler definition to something like:

```ts
const prodService = new AWSService

export async function handler(event: SQSEvent, context: Context, service: Service = prodService)
```

Then in tests you can just pass in the thing you want. We can actually get more control over when `prodService` gets instantiated (i.e. we can import the handler function without instantiating all the things it needs) if we then have a separate file to point the Lambda at which just contains a thin wrapper for the `handler` function and the necessary instantiations to run it, but we can do that as a separate piece of work

src/services/db-client.ts
```diff
@@ -0,0 +1,15 @@
+import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
+import tracer from '../commons/tracer';
+import { AppConfigService } from './app-config-service';
+import { NodeHttpHandler } from '@smithy/node-http-handler';
+export { DynamoDBClient as DbClient } from '@aws-sdk/client-dynamodb';
+
+const dbClient = tracer.captureAWSv3Client(
```
ede-gds: **Suggestion:** We can give ourselves more control over when this construction happens:

```
let dbClient: DynamoDBClient | undefined;

export function getDBClient(): DynamoDBClient {
  dbClient ??= new DynamoDBClient(...);
  return dbClient;
}
```

src/tables/intervention-events.ts
```diff
@@ -0,0 +1,54 @@
...
+  componentId: z.string(),
+  originatingComponentId: z.string().optional(),
+  requesterId: z.string().optional(),
+  originatorReferenceId: z.union([z.string(), z.array(z.string())]).optional(),
+  ttl: z.number().optional(),
+});
+
+export type InterventionEvent = z.infer<typeof schema>;
+
+export class InterventionEventsTable extends DbTableService<typeof schema> {
```
ede-gds: **Suggestion:** If we use composition for wrangling this stuff this class stays pretty much the same except we have something like:

```
export class InterventionEventsTable {
  constructor(
    private recordService: RecordService<typeof schema>
  ) {}

  ...
  async fetchEventsForAccount(accountId: string) {
    return this.recordService.queryAndValidate({
    ...
  }
}
```

This way we can explicitly pass in a fake record service at test time. We might consider a little caching layer for the record services to reuse them too?

src/tables/intervention-events.ts
```diff
@@ -0,0 +1,54 @@
...
+    super(appConfig.interventionEventsTableName, dbClient, schema);
+  }
+
+  async fetchEventsForAccount(accountId: string) {
+    return this.queryAndValidate({
+      TableName: this.tableName,
+      KeyConditionExpression: '#accountId = :account_value',
+      ExpressionAttributeNames: { '#accountId': 'accountId' },
+      ExpressionAttributeValues: { ':account_value': { S: accountId } },
+    });
```
ede-gds: **Thought:** I wonder if we can make the interface here more friendly so this class doesn't have to care about the Dynamo query syntax

ede-gds (commented):
Thanks for this (and sorry for the volume of comments) - probably worth talking through this today because there's a bunch of stuff here that's been in my head and I probably haven't externalised properly

src/services/db-table-service.ts
edlouth: Done

src/services/db-client.ts
edlouth: Done

src/tables/intervention-events.ts
edlouth: Done

src/services/db-table-service.ts
edlouth: Done

src/tables/intervention-events.ts
edlouth: Done

src/tables/intervention-events.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const appConfig = AppConfigService.getInstance();
+
+export enum InterventionState {
+  ACTIVE = 'ACTIVE',
+  IGNORED = 'IGNORED',
+  SUPERSEDED = 'SUPERSEDED',
+  MITIGATED = 'MITIGATED',
+  REMOVED = 'REMOVED',
+}
```
ede-gds, line 15: **Thought:** Maybe we should leave superseded out for now? Because (I think) we have no way to set it currently, and based on what we talked about around thinking about superseding as mitigation by the fraud pod

src/services/db-table-service.ts
```diff
@@ -0,0 +1,46 @@
...
+import TableConfig from '../tables/table-config';
+
+type ArrayFromT<T extends ZodObject<ZodRawShape>> = z.infer<z.ZodArray<T>>;
+
+export interface RecordService<T extends ZodObject<ZodRawShape>> {
+  queryByPkAndValidate(partionKeyValue: string): Promise<ArrayFromT<T>>;
+  batchWrite(input: ArrayFromT<T>): Promise<void>;
+}
+
+export class DbTableService<T extends ZodObject<ZodRawShape> = ZodObject<ZodRawShape>> implements RecordService<T> {
```
ede-gds: **nitpick:**

```suggestion
export class DynamoDBRecordService<T extends ZodObject<ZodRawShape> = ZodObject<ZodRawShape>> implements RecordService<T> {
```

src/handlers/tests/interventions-processor-handler.test.ts
```diff
@@ -151,15 +152,15 @@ describe('intervention processor handler', () => {
   describe('handle', () => {
     it('does nothing if SQS event contains no record', async () => {
       const loggerWarnSpy = vi.spyOn(logger, 'warn');
-      await handler({ Records: [] }, mockContext);
+      await handler({ Records: [] }, mockContext, new InterventionEventsServiceMemory([]));
```
ede-gds: **Suggestion (non-blocking):** We can probably do something like `const emptyInterventionEventsService = new InterventionEventsServiceMemory([])` at the top of this file and use it everywhere

src/tables/intervention-events.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+export class InterventionEventsTable implements InterventionEventsService {
+  constructor(private readonly recordService: RecordService<typeof schema>) {}
+
+  fetchEventsForAccount(accountId: string) {
+    return this.recordService.queryByPkAndValidate(accountId);
+  }
+}
```
ede-gds: **Nitpick:** my personal preference is something like:

```suggestion
export class InMemoryInterventionEventsService implements InterventionEventsService {
  constructor(readonly events: InterventionEvent[]) {}

  fetchEventsForAccount() {
    return Promise.resolve(this.events);
  }
}

export class PersistentInterventionEventsService implements InterventionEventsService {
  constructor(private readonly recordService: RecordService<typeof schema>) {}

  fetchEventsForAccount(accountId: string) {
    return this.recordService.queryByPkAndValidate(accountId);
  }
}
```

ede-gds, line 66: **Suggestion:** We should add a (very simple) test suite for the "real" class (which means we probably have to add a test double that implements `RecordService`)

ede-gds (commented):
Looks great, just a few minor things

src/handlers/tests/interventions-processor-handler.test.ts
edlouth: Done

src/services/db-table-service.ts
edlouth: Done

src/tables/intervention-events.ts
edlouth: Done

edlouth: Done

src/tables/intervention-events.ts
edlouth: `SUPERSEDED` is more like `REMOVED`, it is currently used when a Fraud user removes `TEMPORARY_SUSPENSION` and adds `RESET_PASSWORD` for example.

src/services/test/dynamo-db-record-service.test.ts
```diff
@@ -0,0 +1,115 @@
...
+});
+
+const tableConfig: TableConfig<typeof schema> = {
+  tableName: 'test-table',
+  partitionKeyName: 'pk1',
+  schema,
+};
+
+beforeEach(() => {
+  vi.clearAllMocks();
```
ede-gds: **Nitpick:** I think we can get rid of this because we don't have any vitest-managed mocks in here?

ede-gds (approved):
Looks great, thanks!

src/services/test/dynamo-db-record-service.test.ts
edlouth: I have changed this to `ddbMock.reset();`, I don't think it is really required though.

PR #967, opened 2026-06-02
FPAD-8124: Intervention events table

edlouth:
## What

Create InterventionEvents DynamoDB table in Terraform

## Notes

I have deployed this in the `dev` environment to test

### Issue tracking

- [FPAD-8124](https://govukverify.atlassian.net/browse/FPAD-8124)

[FPAD-8124]: https://govukverify.atlassian.net/browse/FPAD-8124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/infra/core/samconfig.toml
+++ b/src/infra/core/samconfig.toml
@@ -6,5 +6,8 @@ s3_prefix = "ais-core"
 region = "eu-west-2"
 confirm_changeset = true
 capabilities = "CAPABILITY_IAM"
-parameter_overrides = "Environment=\"dev\""
+parameter_overrides = "Environment=\"dev\" ProductTagValue=\"GOV.UK One Login\" SystemTagValue=\"Account Interventions Service\" OwnerTagValue=\"interventions@digital.cabinet-office.gov.uk\" SourceTagValue=\"govuk-one-login/account-interventions-service/src/infra/core/template.yaml\""
 image_repositories = []
+
+[default.global.parameters]
+region = "eu-west-2"
--- a/src/infra/main/template.yaml
+++ b/src/infra/main/template.yaml
@@ -226,6 +226,8 @@ Resources:
           METRIC_SERVICE_NAME: 'InterventionsProcessorFunction'
           TABLE_NAME:
             Fn::ImportValue: !Sub '${CoreStackName}-AccountStatusTableName'
+          INTERVENTION_EVENTS_TABLE_NAME:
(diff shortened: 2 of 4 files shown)


src/infra/core/samconfig.toml
```diff
@@ -6,5 +6,8 @@ s3_prefix = "ais-core"
 region = "eu-west-2"
 confirm_changeset = true
 capabilities = "CAPABILITY_IAM"
-parameter_overrides = "Environment=\"dev\""
+parameter_overrides = "Environment=\"dev\" ProductTagValue=\"GOV.UK One Login\" SystemTagValue=\"Account Interventions Service\" OwnerTagValue=\"interventions@digital.cabinet-office.gov.uk\" SourceTagValue=\"govuk-one-login/account-interventions-service/src/infra/core/template.yaml\""
```
kirstyallen1, line 9: Question: Were these parameter overrides not being correctly applied in the tags for the AccountStatus table before being added here?

edlouth: Normally this template is deployed by a pipeline, so the tags are set as part of that.  They can also be set with `sam deploy --guided`.  This is just storing the default values per say, so not really required, but it just means I can be more confident running `sam deploy` without the `--guided` flag.

src/infra/main/template.yaml
```diff
@@ -331,6 +333,14 @@ Resources:
               - 'dynamodb:Query'
             Resource:
               - Fn::ImportValue: !Sub '${CoreStackName}-AccountStatusTableArn'
+          - Effect: Allow
+            Action:
+              - 'dynamodb:PutItem'
+              - 'dynamodb:BatchWriteItem'
+              - 'dynamodb:UpdateItem'
+              - 'dynamodb:Query'
```
ede-gds, line 339: **Suggestion:** Nitpicky but we don't need all these right now, right? probably better to keep this as a minimal set for security, and also to help us verify our assumptions about the system

ede-gds (approved):
lgtm, one minor thing

src/infra/main/template.yaml
edlouth: I probably copied these from elsewhere in the template.
I think we need `Query` and `BatchWriteItem` for now, I can remove the others.

PR #968, opened 2026-06-03
FPAD-8126: Persist interventions model

edlouth:
## What

Persist interventions model

## Notes

This logs out the intervention event rows to add, that will be done in a subsequent ticket.

This is based on https://github.com/govuk-one-login/account-interventions-service/pull/967 and https://github.com/govuk-one-login/account-interventions-service/pull/966.

This doesn't currently handle `IGNORED` events.

This doesn't have much validation.

I need to check the config actually maps to what is expected around `REMOVED` vs `SUPERSEDED`.

### Issue tracking

- [FPAD-8126](https://govukverify.atlassian.net/browse/FPAD-8126)

[FPAD-8126]: https://govukverify.atlassian.net/browse/FPAD-8126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/active-interventions-service.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  return existingInterventionEvents;
+}
+
+/**
+ * Map the state from the account status table to a list of intervention names
+ * @param previousState - Previous state from the account status table
+ * @returns
+ */
+export function perviousStateToInterventions(previousState?: StateDetails): InterventionName[] {
```
kirstyallen1: typo: pervious

edlouth: Fixed

src/services/active-interventions-service.ts
```diff
@@ -0,0 +1,91 @@
...
+  previousState: StateDetails | undefined,
+): Promise<InterventionName[]> {
+  const interventionsFromEvents = await getActiveInterventionsFromEvents(accountId, interventionEventsService);
+
+  const interventionsFromPreviousState = previousStateToInterventions(previousState);
+
+  return combineSets(interventionsFromEvents, interventionsFromPreviousState);
+}
+
+export default getActiveInterventions;
```
kirstyallen1, line 26: Question: Is there a specific reason this is a default export instead of a named export?

edlouth: A good question, I made it the default export as it is the main way into this set of functions, the other exports are only for testing.

kirstyallen1: Thanks, thought so but wanted to check :)

src/services/test/persist-intervention-events.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  extensions: {
+    intervention: {
+      intervention_code: '01',
+      intervention_reason: 'reason',
+    },
+  },
+};
+
+describe('persistInterventionEvents', () => {
+  test('happy path', async () => {
```
kirstyallen1: note: definitely non-blocking, but just wondered if we should be consistent with use of it/test? Can see we tend to use 'it' in other test files I've seen throughout the repo. It might also be helpful for the test cases to be a little more descriptive in what the expected outcome is as this could improve readability/understanding at a glance

kirstyallen1 (approved):

src/services/test/persist-intervention-events.test.ts
edlouth: Fixed

PR #973, opened 2026-06-05
FPAD-8166: Command to send TxMA messages

edlouth:
## What

Command to send TxMA messages

## Why

Makes testing easier

### Issue tracking

- [FPAD-8166](https://govukverify.atlassian.net/browse/FPAD-8166)

[FPAD-8166]: https://govukverify.atlassian.net/browse/FPAD-8166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/commands/send-audit-event.ts
```diff
@@ -0,0 +1,126 @@
...
+    interventionCode: { type: 'string' },
+  },
+});
+
+const isValidEvent = (value: string | undefined): value is EventType =>
+  Object.values(EventType).includes(value as EventType);
+
+// eslint-disable-next-line unicorn/prefer-top-level-await
+void (async () => {
+  const queueUrl = values.queueUrl ?? 'https://sqs.eu-west-2.amazonaws.com/484907510598/ais-main-TxMAIngressQueue';
```
kirstyallen1, line 112: Question: Do we want to have this queue url in plain text here or would it be better in an env variable or similar?

edlouth: I don't think this is particularly secret and we don't currently have an `.env` file or similar to easily put this in.  But I could be persuaded to put this somewhere.

kirstyallen1: Happy with it being left as-is if preferred and there are no objections

kirstyallen1 (approved):

PR #978, opened 2026-06-09
Add new ADRs

ede-gds:
## What

Add a couple of new ADRs

## Why

These codify some approaches we've been following implicitly


edlouth (approved):
Looks good to me

PR #985, opened 2026-06-10
FPAD-8109: Status API v2

edlouth:
## What

Status API v2

## Notes

This has been implemented under `main/v2/ais/{userId}` on a new API Gateway, as it isn't possible to have different API specs for different stage names.

This has been tested in AWS on `dev`.

### Issue tracking

- [FPAD-8109](https://govukverify.atlassian.net/browse/FPAD-8109)

[FPAD-8109]: https://govukverify.atlassian.net/browse/FPAD-8109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
+++ b/feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
@@ -69,6 +69,27 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, BeforeEachScenario }) => 
     },
   );
 
+  Scenario('Happy Path - Get Request to /v2/ais/userId Returns Expected Data', ({ Given, When, Then }) => {
+    Given('I send an intervention message to the TxMA ingress SQS queue', async () => {
+      await sendSQSEvent(testUserId, 'suspendNoAction');
+    });
+
+    When('I invoke an API v2 to retrieve the intervention status of the account', async () => {
+      await timeDelayForTestEnvironment(4000);
+      response = await invokeGetAccountState(testUserId, false, true);
+    });
+
+    Then('I expect the API to return a list of active interventions', () => {
+      expect(response).toEqual({
+        interventions: [
+          {
+            name: 'TEMPORARY_SUSPENSION',
+          },
+        ],
+      });
+    });
+  });
+
   ScenarioOutline(
     'Happy Path - Get Request to /ais/userId with enhanced fields- Returns Expected Data for <aisEventType>',
     ({ Given, When, Then }, { enhancedAisEventType }) => {
--- a/src/handlers/invoke-private-api.ts
+++ b/src/handlers/invoke-private-api.ts
@@ -3,16 +3,17 @@
 import getEnvOrThrow from '../commons/get-env-or-throw';
 import logger from '../commons/logger';
 
-interface CustomEvent {
+export interface CustomEvent {
   userId?: string;
   queryParameters?: string;
   baseUrl?: string;
   endpoint?: string;
   headers?: Record<string, string>;
+  privateApi?: boolean;
 }
 
 export async function handle(event: CustomEvent) {
-  const baseUrl = event.baseUrl ?? getBaseUrl();
+  const baseUrl = event.baseUrl ?? getBaseUrl(event);
   const endpoint = event.endpoint ?? getEndpoint();
   const userId = encodeURI(event.userId ?? getUserId());
   const headers = event.headers ?? { 'Content-Type': 'application/json' };
@@ -42,8 +43,8 @@ function getUserId() {
   return getEnvOrThrow('USER_ID');
 }
 
-function getBaseUrl() {
-  return validateConfiguration('BASE_URL');
+function getBaseUrl(event: CustomEvent) {
+  return event.privateApi ? validateConfiguration('PRIVATE_API_URL') : validateConfiguration('BASE_URL');
 }
 
 function getEndpoint() {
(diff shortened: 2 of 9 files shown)


edlouth:
The feature tests will need fixing to run on AWS

src/infra/main/template.yaml
```diff
@@ -1423,6 +1436,11 @@ Resources:
             - - https://
               - !GetAtt PrivateRestApi.RestApiId
               - !Sub .execute-api.${AWS::Region}.amazonaws.com/${ApiStageName}
+          MAIN_URL: !Join
```
ede-gds: **Suggestion:** I know we briefly called this the main gateway but now that it's called `PrivateApi` (?) can we make this name a bit more descriptive?

src/handlers/status-retriever-handler.ts
```diff
@@ -40,44 +38,84 @@ export async function handle(event: APIGatewayEvent, context: Context): Promise<
...
-    const accountStatus = transformResponseFromDynamoDatabase(response);
+  metric.publishStoredMetrics();
+
+  return {
+    statusCode: 500,
+    body: JSON.stringify({ message: 'Internal Server Error.' }),
+  };
+}
+
+async function v1Handler(userId: string, historyQuery: string | undefined) {
```
ede-gds: **Nitpick:** We could be more descriptive with these function names - `vNStatusApi` or something?

src/handlers/invoke-private-api.ts
```diff
@@ -3,16 +3,17 @@
...
 import logger from '../commons/logger';
 
-interface CustomEvent {
+export interface CustomEvent {
   userId?: string;
   queryParameters?: string;
   baseUrl?: string;
   endpoint?: string;
   headers?: Record<string, string>;
+  mainGateway?: boolean;
```
ede-gds: **Suggestion:** A boolean is a bit limiting here - can we use `apiVersion` or something instead with the type as a union of string literals?

src/handlers/tests/status-retriever-handler.test.ts
```diff
@@ -574,3 +581,112 @@ describe('status-retriever-handler', () => {
...
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date(Date.UTC(2023, 4, 30)).getTime());
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+    vi.resetAllMocks();
+  });
+
+  test('v2 no account found', async () => {
```
ede-gds: **Issue:** It's a small thing but can we make these test names more descriptive? Mainly so that if someone breaks them they have a head start on what's actually broken. I think using `it` instead of `test` is helpful for this but I don't feel strongly.

For this one something like `it('returns an empty list of interventions if the requested account doesn't exist'`

src/handlers/tests/status-retriever-handler.test.ts
```diff
@@ -574,3 +581,112 @@ describe('status-retriever-handler', () => {
     expect(addMetric).toHaveBeenCalled();
   });
 });
+
+describe('status-retriever-handler v2', () => {
```
ede-gds: **Suggestion (non-blocking):** Very minor but I'd consider making this more human-understandable - something like `describe('v2 Status API handler'`

feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
```diff
@@ -69,6 +69,27 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, BeforeEachScenario }) =>
...
+    Given('I send an intervention message to the TxMA ingress SQS queue', async () => {
+      await sendSQSEvent(testUserId, 'suspendNoAction');
+    });
+
+    When('I invoke an API v2 to retrieve the intervention status of the account', async () => {
+      await timeDelayForTestEnvironment(4000);
+      response = await invokeGetAccountState(testUserId, false, true);
+    });
+
+    Then('I expect the output to be as expected', () => {
```
ede-gds: **Suggestion (non-blocking):** As elsewhere, I think this could be more useful

```suggestion
    Then('I expect the API to return a list of active interventions, () => {
```

ede-gds (approved):
The handler looks great, just minor bits

src/handlers/invoke-private-api.ts
edlouth: Given this is totally within our gift to change, I think I will leave this as a `boolean` for now. It doesn't neatly map to `apiVersion` in terms of `v1`/`v2`.

src/handlers/status-retriever-handler.ts
edlouth: Done

src/infra/main/template.yaml
edlouth: I have renamed this `PRIVATE_API_URL` but really these names are all a bit confusing, any suggestions for something better welcome.

src/handlers/invoke-private-api.ts
ede-gds: Very fair, sounds good

src/infra/main/template.yaml
ede-gds: agreed - I suppose the only suggestion I have is making it very clear that the old one is old and will be deprecated at some point

feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
edlouth: Done

PR #986, opened 2026-06-11
Feature-tests: test for batch failures

edlouth:
## Notes

I haven't yet checked that this would have caught our issue.

### DIFF
--- a/feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
+++ b/feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
@@ -17,7 +17,6 @@ import {
 import { aisEventResponse, AisResponseType } from '../../../utils/ais-events-responses';
 import { updateItemInTable, getRecordFromTable } from '../../../utils/dynamo-database-methods';
 import { cloudwatchLogs, LogEvent } from '../../../utils/cloudwatch-logs-service';
-import { aisEventsWithEnhancedFields } from '../../../utils/enhanced-ais-events';
 
 const feature = await loadFeature('./tests/resources/features/aisGET/InvokeApiGateWay-HappyPath.feature');
 
@@ -39,7 +38,7 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, BeforeEachScenario }) => 
 
   ScenarioOutline(
     'Happy Path - Get Request to /ais/userId - Returns Expected allowable intervention codes from Egress Queue <aisEventType>',
-    ({ Given, When, Then }, { aisEventType, historyValue }) => {
+    ({ Given, When, Then, And }, { aisEventType, historyValue }) => {
       Given('I send an <aisEventType> intervention to the TxMA ingress SQS queue', async () => {
         await sendSQSEvent(testUserId, aisEventType);
       });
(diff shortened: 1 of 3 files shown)


feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
```diff
@@ -37,9 +36,9 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, BeforeEachScenario }) =>
     testUserId = generateRandomTestUserId();
   });
 
-  ScenarioOutline(
+  ScenarioOutline.only(
```
kirstyallen1: Suggestion: remove .only

edlouth: Good catch

edlouth: Fixed

kirstyallen1 (approved):

ede-gds:
@edlouth can you write a description of the underlying issue in the commit please? (I know this might not fix it)

edlouth:
> @edlouth can you write a description of the underlying issue in the commit please? (I know this might not fix it)

I have added a further description to the commit message.  This PR should possibly be in draft, I haven't yet confirmed that it works as expected, and potentially a better feature test would test the expected behaviour, now that we are saving Intervention Events to the database.

ede-gds:
> > @edlouth can you write a description of the underlying issue in the commit please? (I know this might not fix it)
> 
> I have added a further description to the commit message. This PR should possibly be in draft, I haven't yet confirmed that it works as expected, and potentially a better feature test would test the expected behaviour, now that we are saving Intervention Events to the database.

Thanks, understood

feature-tests/tests/resources/step-definitions/invoke-api-gateway-happy-path.step.ts
```diff
@@ -66,6 +65,12 @@ describeFeature(feature, ({ Scenario, ScenarioOutline, BeforeEachScenario }) =>
           }
         },
       );
+
+      And('I expect the intervention processor not have any batch item failures', () => {
+        const hasRetries = cloudwatchLogs.hasRetries();
```
ede-gds, line 70: **Thought:** It feels like there should be a better way to check this than looking in the logs

ede-gds (commented):
As per our slack conversation, this looks fine but it's probably best to figure out what the actual artifact to test is instead of checking the logs, and writing some proper tests for the persist interventions work is probably the right time to do that

ede-gds:
@edlouth can we close this and do something like move whatever we've learnt to a new ticket?

edlouth:
> @edlouth can we close this and do something like move whatever we've learnt to a new ticket?

Yes.  The answer was to switch to explicit feature tests for each new feature, ideally as part of the main ticket.

ede-gds:
> > @edlouth can we close this and do something like move whatever we've learnt to a new ticket?
> 
> Yes. The answer was to switch to explicit feature tests for each new feature, ideally as part of the main ticket.

I guess we could also do with a general solution for checking that a lambda completed successfully?

edlouth:
> > > @edlouth can we close this and do something like move whatever we've learnt to a new ticket?
> > 
> > 
> > Yes. The answer was to switch to explicit feature tests for each new feature, ideally as part of the main ticket.
> 
> I guess we could also do with a general solution for checking that a lambda completed successfully?

That would be good, however it is hard to know how to do this, beyond checking for this comment.

PR #995, opened 2026-06-16
Fix feature tests lock file

edlouth:
## What

Fix feature tests lock file

## Why

Probably broken as part of a merge conflict


ede-gds (approved):
lgtm (assuming you've run `npm install` locally)

PR #997, opened 2026-06-16
Remove EndpointConfiguration from PrivateAPI

edlouth:
## What

Remove EndpointConfiguration from PrivateAPI

## Why

The first endpoint in staging can't be found, which is blocking deployment

## Notes

At some point we should put this back, so that other teams have access to this Gateway

### DIFF
--- a/src/infra/main/template.yaml
+++ b/src/infra/main/template.yaml
@@ -1687,10 +1687,12 @@ Resources:
       TracingEnabled: true
       EndpointConfiguration:
         Type: PRIVATE
-        VpcEndpointIds: !If
-          - IsNotInternal
-          - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
-          - []
+        # TODO - figure out status of the perf VPC id
+        VpcEndpointIds: []
+        # VpcEndpointIds: !If
+        #   - IsNotInternal
+        #   - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
+        #   - []
       Tags:
         CheckovRulesToSkip: CKV_AWS_120
         Product: !Ref ProductTagValue


src/infra/main/template.yaml
```diff
@@ -1685,12 +1685,12 @@ Resources:
...
-        VpcEndpointIds: !If
-          - IsNotInternal
-          - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
-          - []
+      # EndpointConfiguration:
+      #   Type: PRIVATE
+      #   VpcEndpointIds: !If
+      #     - IsNotInternal
+      #     - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
+      #     - []
```
ede-gds: **Issue:** I'm (probably incorrectly) worried about the default endpoint status if we don't explicitly set the type to `PRIVATE` - can we just do this instead?

```suggestion
      EndpointConfiguration:
        Type: PRIVATE
        VpcEndpointIds: []
      # TODO - figure out status of the perf VPC id
      #   VpcEndpointIds: !If
      #     - IsNotInternal
      #     - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
      #     - []
```

edlouth: Makes sense to me

src/infra/main/template.yaml
```diff
@@ -1687,10 +1687,11 @@ Resources:
       TracingEnabled: true
       EndpointConfiguration:
         Type: PRIVATE
-        VpcEndpointIds: !If
-          - IsNotInternal
-          - !Split [',', !FindInMap [EnvConfig, !Ref Environment, UniqueVPCEndpointIDs]]
-          - []
+        # TODO - figure out status of the perf VPC id
```
ede-gds, line 1690: **Nitpick:** Somehow this line got nuked from my suggestion

```suggestion
        VpcEndpointIds: []
        # TODO - figure out status of the perf VPC id
```

edlouth: Now with `VpcEndpointIds: []`

ede-gds (approved):

PR #1001, opened 2026-06-18
FPAD-8173: Persist an intervention event to the db when a transition is ignored

kirstyallen1:
## What
Write to the database when an intervention transition is ignored

## Why
We want to record when intervention transitions are ignored

### Issue tracking

- [FPAD-8173](https://govukverify.atlassian.net/browse/FPAD-8173)

[FPAD-8173]: https://govukverify.atlassian.net/browse/FPAD-8173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/tests/interventions-processor-handler.test.ts
```diff
@@ -217,6 +217,56 @@ describe('intervention processor handler', () => {
       expect(publishTimeToResolveMetrics).not.toHaveBeenCalled();
     });
 
+    it.only('should persist an ignored intervention event to the database when a StateTransitionError is received', async () => {
```
edlouth:
```suggestion
    it('should persist an ignored intervention event to the database when a StateTransitionError is received', async () => {
```

src/services/persist-intervention-events.ts
```diff
@@ -160,7 +160,14 @@ const config: Record<EventsEnum, InterventionUpdate[]> = {
...
-    { interventionName: InterventionName.PERMANENT_SUSPENSION, interventionState: InterventionState.REMOVED },
+    { interventionName: InterventionName.PERMENANT_SUSPENSION, interventionState: InterventionState.REMOVED },
+  ],
+
+  [EventsEnum.INTERVENTION_EVENT_IGNORED]: [
+    { interventionName: InterventionName.TEMPORARY_SUSPENSION, interventionState: InterventionState.IGNORED },
+    { interventionName: InterventionName.PERMENANT_SUSPENSION, interventionState: InterventionState.IGNORED },
+    { interventionName: InterventionName.RESET_PASSWORD, interventionState: InterventionState.IGNORED },
+    { interventionName: InterventionName.REPROVE_IDENTITY, interventionState: InterventionState.IGNORED },
   ],
```
edlouth: I think we should only include the actual intervention requested

kirstyallen1: How exactly do you mean here?

edlouth: If the user currently has `RESET_PASSWORD` and `REPROVE_IDENTITY` interventions on their account, if the TICF user tried to add `RESET_PASSWORD` again, we want a single record inserting into the database table for `RESET_PASSWORD` with a state of `IGNORED`.  I think your code will insert two rows.

edlouth (approved):

edlouth:
@kirstyallen1 if you could rebase, this should be good to merge.

The rebase will bring in a new major version of eslint rules, so you may need to fix some lint issues.

kirstyallen1:
> @kirstyallen1 if you could rebase, this should be good to merge.
> 
> The rebase will bring in a new major version of eslint rules, so you may need to fix some lint issues.

Ah, I must have rebased just before this - will rebase again. Thanks for the heads up @edlouth :)

PR #1005, opened 2026-06-18
FPAD-8262: Persist interventions feature tests

edlouth:
## What

Persist interventions feature tests

## Why

Catch issues at build time rather than in production

### Issue tracking

- [FPAD-8262](https://govukverify.atlassian.net/browse/FPAD-8262)

[FPAD-8262]: https://govukverify.atlassian.net/browse/FPAD-8262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature-tests/apiEndpoints/endpoints.ts
+++ b/feature-tests/apiEndpoints/endpoints.ts
@@ -22,4 +22,5 @@ export default class EndPoints {
   public static PATH_AIS = '/ais/';
   public static INVOKE_PRIVATE_API_GATEWAY = `${process.env.SAM_STACK_NAME}-InvokePrivateAPIGatewayFunction`;
   public static TABLE_NAME = 'ais-core-account-status';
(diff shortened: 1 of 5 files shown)


kirstyallen1:
Looks good @edlouth 👍 Quick question - the ticket mentioned one or more feature tests as is on here, do we need further tickets for other scenarios or do we intend to just add as we go?

kirstyallen1 (approved):

edlouth:
> Looks good @edlouth 👍 Quick question - the ticket mentioned one or more feature tests as is on here, do we need further tickets for other scenarios or do we intend to just add as we go?

I have added two feature tests.  These are basically smoke tests, I am not aiming to cover all scenarios, that is what unit tests are for.  For now I think this is sufficient, if we add more functionality to the endpoint, or more complexity to the state machine, we should potentially add some more feature tests.

PR #1009, opened 2026-06-19
FPAD-8283: Account status typed table service

edlouth:
## What

Account status typed table service

## Why

Refactor gives better validation on DynamoDB table

Easier testing without mocking

### Issue tracking

- [FPAD-8283](https://govukverify.atlassian.net/browse/FPAD-8283)

[FPAD-8283]: https://govukverify.atlassian.net/browse/FPAD-8283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tables/account-status.ts
```diff
@@ -0,0 +1,154 @@
...
+    return this.recordService.update(accountId, {
+      UpdateExpression: 'SET #isAccountDeleted = :isAccountDeleted, #ttl = :ttl, #deletedAt = :deletedAt',
+      ExpressionAttributeNames: {
+        '#isAccountDeleted': 'isAccountDeleted',
+        '#ttl': 'ttl',
+        '#deletedAt': 'deletedAt',
+      },
+      ExpressionAttributeValues: {
+        ':isAccountDeleted': true,
+        ':ttl': ttl.toString(),
```
kirstyallen1: Question: Does this have the potential down the line (meaning in a later call) for a zod error to be thrown if the update converts the ttl to a string but the zod schema expects a number, and this is returned as a string if/when the account info is later retrieved?

src/tables/account-status.ts
```diff
@@ -0,0 +1,154 @@
...
+        ':false': false,
+        ':deletedAt': now.milliseconds.toString(),
```
kirstyallen1: Same question as above

src/services/dynamo-db-record-service.ts
```diff
@@ -1,12 +1,33 @@
...
+  UpdateCommandOutput,
+} from '@aws-sdk/lib-dynamodb';
 import TableConfig from '../tables/table-config';
 
 type ArrayFromT<T extends ZodObject<ZodRawShape>> = z.infer<z.ZodArray<T>>;
+type PickedArray<T extends ZodObject<ZodRawShape>, K extends keyof z.infer<T>> = Pick<z.infer<T>, K>[];
 
 export interface RecordService<T extends ZodObject<ZodRawShape>> {
-  queryByPkAndValidate(partionKeyValue: string): Promise<ArrayFromT<T>>;
+  queryByPkAndValidate(partionKeyValue: string, includedKeys?: undefined): Promise<ArrayFromT<T>>;
```
kirstyallen1: Typo: Looks like this was already present pre-refactor but this file has `partion` used throughout instead of `partition`

edlouth: Good catch

src/tables/account-status.ts
edlouth: This is a good question, the types seem to work without `.toString()`, so I have changed them

kirstyallen1 (approved):

src/tables/account-status.ts
```diff
@@ -0,0 +1,154 @@
...
+      ExpressionAttributeValues: {
+        ':isAccountDeleted': true,
+        ':ttl': ttl,
+        ':false': false,
+        ':deletedAt': now.milliseconds,
+      },
+      ReturnValues: 'ALL_NEW',
+      ConditionExpression:
+        'attribute_exists(pk) AND (attribute_not_exists(isAccountDeleted) OR isAccountDeleted = :false)',
+    });
```
ede-gds, line 149: **Suggestion:** This is a custom dynamo query so let's add unit tests that use a record service with a dynamo mock (and verify that the dynamo mock actually checks whether you're sending it legitimate dynamo queries)

edlouth: I have written a test as far as `mockClient(DynamoDBDocumentClient)`

PR #1028, opened 2026-06-26
FPAD-8384: UI Base

edlouth:
## What

UI Base on Fastify with numjusks and govuk-frontend

### Issue tracking

- [FPAD-8384](https://govukverify.atlassian.net/browse/FPAD-8384)

[FPAD-8384]: https://govukverify.atlassian.net/browse/FPAD-8384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -9,4 +9,4 @@ sonar.tests=src/
 sonar.test.inclusions=**/*.test.ts
 sonar.sourceEncoding=UTF-8
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts
+sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts, src/frontend/local.ts, src/handlers/frontend-handler.ts
--- a/src/frontend/app.ts
+++ b/src/frontend/app.ts
@@ -1,0 +1,30 @@
+import fastify from 'fastify';
+import view from '@fastify/view';
+import staticFiles from '@fastify/static';
+import nunjucks from 'nunjucks';
+import path from 'node:path';
+
+export function init() {
+  const server = fastify();
+
+  server.register(staticFiles, {
+    root: path.join(__dirname, '../../node_modules/govuk-frontend/dist/govuk/assets'),
+    prefix: '/assets/',
+  });
+
+  // Serve govuk-frontend CSS and JS directly under /assets/
+  const govukDistribution = path.join(__dirname, '../../node_modules/govuk-frontend/dist/govuk');
+
+  for (const file of ['govuk-frontend.min.css', 'govuk-frontend.min.js']) {
+    server.get(`/assets/${file}`, (_request, reply) => reply.sendFile(file, govukDistribution));
+  }
+
+  server.register(view, {
+    engine: { nunjucks },
+    templates: ['src/frontend/views', 'node_modules/govuk-frontend/dist'],
+  });
+
+  server.get('/', async (_request, reply) => reply.view('index.njk'));
+
+  return server;
+}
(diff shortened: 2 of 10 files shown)


src/frontend/app.ts
```diff
@@ -0,0 +1,32 @@
...
+    engine: { nunjucks },
+    templates: ['src/frontend/views', 'node_modules/govuk-frontend/dist'],
+  });
+
+  server.get('/', async (_request, reply) => reply.view('index.njk'));
+
+  return server;
+}
+
+/* istanbul ignore next */
```
ede-gds: **Issue:** Ignores should have a justification

**Suggestion:** This makes me itch and I don't think we need it - if we want to invoke locally we can just have a script that calls L28 directly and if we want to run it for testing we can do similar in the test setup

src/handlers/tests/frontend-handler.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const mockProxy = vi.fn().mockResolvedValue({ statusCode: 200 });
+vi.mock('@fastify/aws-lambda', () => ({ default: () => mockProxy }));
+vi.mock('../../frontend/app', () => ({ init: vi.fn().mockReturnValue({}) }));
+
+describe('frontend-handler', () => {
+  it('exports a handler function', async () => {
+    const { handler } = await import('../frontend-handler');
+    expect(typeof handler).toBe('function');
+  });
```
ede-gds: **Suggestion (non-blocking):** We can probably get rid of this - I get why it exists but this stuff is also checked by the other test, and the lambda integration should be checked by whatever post-merge integration smoke test things we end up with.

ParkRanger-Pro:
Has there been an active decision on to use the same repo of interface work?

edlouth:
> Has there been an active decision on to use the same repo of interface work?

Yes and no. I have made the decision, on the basis that we can create a separate GitHub repository later, if the need arises.  Happy for the team to discuss though.

src/handlers/tests/frontend-handler.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  it('delegates to the awsLambdaFastify proxy', async () => {
```
ede-gds: **Thought:** I'm conflicted about this - I actually think the logic in `frontend-handler.ts` is simple enough that it doesn't need testing (aside from whatever feature tests we eventually run on the actual Lambdas)

ede-gds (commented):
Thanks, a few comments - let me know if you want to talk about anything

sonar-project.properties
```diff
@@ -9,4 +9,4 @@ sonar.tests=src/
 sonar.test.inclusions=**/*.test.ts
 sonar.sourceEncoding=UTF-8
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts
+sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts, src/frontend/local.ts, src/handlers/frontend-handler.ts
```
ede-gds, line 12: **Suggestion:** Sorry to be boring - can you exclude this in code using `/* istanbul ignore file */` with a justification comment? like `production only` or something

ede-gds (approved):
lgtm, just one minor thing

sonar-project.properties
edlouth: Sadly `/* istanbul ignore file */` doesn't work as we are using the `v8` provider, I did try.  I am welcome to any suggestions on how to get this to work.

ede-gds: grrrr

ede-gds: can you just add a comment at the top of the file? and I'll go tinker on a separate branch

PR #1029, opened 2026-06-26
FPAD-8128: Persist interventions TTL

kirstyallen1:
## What
When an intervention has reached a terminal state we want to update its events to have a ttl value so that the data is deleted once the ttl period elapses

## Why
We can and should only store data for a certain period of time

## Notes
This solution assumes that an intervention has reached a terminal state when the intervention state is no longer active. The intervention name is then used to decide which events should be modified to have a ttl value so that only the inactive events for the required intervention are given a ttl value

### Issue tracking

- [FPAD-8128](https://govukverify.atlassian.net/browse/FPAD-8128)

[FPAD-8128]: https://govukverify.atlassian.net/browse/FPAD-8128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/persist-intervention-events.ts
+++ b/src/services/persist-intervention-events.ts
@@ -136,14 +136,14 @@ export function enrichEvents(updates: InterventionUpdate[], message: Interventio
 }
 
 export async function setTtlOnInactiveEvents(
-  accountId: string, 
+  accountId: string,
   interventionEventsService: InterventionEventsService,
   closedInterventionNames: InterventionName[]
 ) {
   const allEvents = await interventionEventsService.fetchEventsForAccount(accountId);
   const now = getCurrentTimestamp();
   const ttl = now.seconds + appConfig.maxRetentionSeconds;
-  
+
   const eventsNeedingTtl = allEvents.filter(
     (event) =>
       closedInterventionNames.includes(event.interventionName) &&
--- a/src/services/persist-intervention-events.ts
+++ b/src/services/persist-intervention-events.ts
@@ -7,11 +7,14 @@ import { InterventionEvent, InterventionEventsService } from '../tables/interven
 import { randomUUID } from 'node:crypto';
 import getActiveInterventions from './active-interventions-service';
 import { InterventionName } from '../data-types/intervention-name';
+import { AppConfigService } from './app-config-service';
 
 interface InterventionUpdate {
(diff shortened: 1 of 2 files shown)


src/services/persist-intervention-events.ts
```diff
@@ -127,6 +135,28 @@ export function enrichEvents(updates: InterventionUpdate[], message: Interventio
...
+  closedInterventionNames: InterventionName[]
+) {
+  const allEvents = await interventionEventsService.fetchEventsForAccount(accountId);
+  const now = getCurrentTimestamp();
+  const ttl = now.seconds + appConfig.maxRetentionSeconds;
+  
+  const eventsNeedingTtl = allEvents.filter(
+    (event) =>
+      closedInterventionNames.includes(event.interventionName) &&
+      event.interventionState !== InterventionState.ACTIVE &&
```
edlouth:
```suggestion
```

We don't need this condition, we also want to set a `ttl` on the rows with state `ACTIVE`

src/services/test/persist-intervention-events.test.ts
```diff
@@ -251,3 +252,153 @@ describe('persistIgnoredInterventionEvent', () => {
...
+        componentId: 'test',
+      },
+    ]);
+    const appendEventsSpy = vi.spyOn(service, 'appendEvents');
+
+    await setTtlOnInactiveEvents('1', service, [InterventionName.TEMPORARY_SUSPENSION]);
+
+    expect(appendEventsSpy).toHaveBeenCalledExactlyOnceWith([
+      expect.objectContaining({ eventId: 'event-2', ttl: EXPECTED_TTL }),
+    ]);
```
edlouth, line 414: I would expect this to include two rows

src/services/persist-intervention-events.ts
kirstyallen1: Amended

src/services/test/persist-intervention-events.test.ts
kirstyallen1: This unit test should reflect this accurately now 🤞

edlouth (approved):
Looks great

PR #1046, opened 2026-07-01
FPAD-8418: Switch to fake timers

kirstyallen1:
## What
Switch to use FakeTimers in tests

## Why
We should have a uniform approach across tests

### Issue tracking

- [FPAD-8418](https://govukverify.atlassian.net/browse/FPAD-8418)

[FPAD-8418]: https://govukverify.atlassian.net/browse/FPAD-8418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handlers/tests/interventions-processor-handler.test.ts
+++ b/src/handlers/tests/interventions-processor-handler.test.ts
@@ -19,22 +19,14 @@ vi.mock('../../commons/metrics');
 vi.mock('../../services/send-audit-events');
 vi.mock('../../commons/metrics-helper');
 
-vi.mock('../../commons/get-current-timestamp', () => ({
-  getCurrentTimestamp: vi.fn().mockImplementation(() => ({
-    milliseconds: 1234567890,
-    isoString: 'today',
-    seconds: 1234567,
-  })),
-}));
+const FIXED_TIME_MS = 1234567890;
+const FIXED_TIME_S = 1234567;
 
-const now = getCurrentTimestamp();
-const t0ms = now.milliseconds;
-const t0s = now.seconds;
 
 const interventionEventBody: InterventionEventMessage = {
   component_id: '',
-  timestamp: t0s - 5,
-  event_timestamp_ms: t0ms - 5000,
+  timestamp: FIXED_TIME_S - 5,
+  event_timestamp_ms: FIXED_TIME_MS - 5000,
   user: {
     user_id: 'abc',
   },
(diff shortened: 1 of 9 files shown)


src/handlers/tests/interventions-processor-handler.test.ts
```diff
@@ -115,6 +113,8 @@ describe('intervention processor handler', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+    vi.useFakeTimers();
+    vi.setSystemTime(1234567890);
```
edlouth:
```suggestion
    vi.setSystemTime(FIXED_TIME_MS);
```

src/services/test/validate-event.test.ts
```diff
@@ -17,29 +17,27 @@ import { TICF_ACCOUNT_INTERVENTION } from '@govuk-one-login/event-catalogue/TICF
...
   isAccountDeleted: false,
   history: [],
   intervention: AISInterventionTypes.AIS_ACCOUNT_UNSUSPENDED,
 };
 
 describe('validateEventAgainstSchema', () => {
+  let timestamp: ReturnType<typeof getCurrentTimestamp>;
+  beforeAll(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(1234567890);
```
edlouth:
```suggestion
    vi.setSystemTime(FIXED_TIME_MS);
```

edlouth (commented):
I think we should be consistent in adding everywhere we use `vi.useFakeTimers();`

```
afterAll(() => {
    vi.useRealTimers();
  });
```

kirstyallen1:
> I think we should be consistent in adding everywhere we use `vi.useFakeTimers();`
> 
> ```
> afterAll(() => {
>     vi.useRealTimers();
>   });
> ```

Done

src/services/test/validate-event.test.ts
kirstyallen1: Done

src/handlers/tests/interventions-processor-handler.test.ts
kirstyallen1: Done

edlouth (commented):
This all looks great.

Can we check for places like this https://github.com/govuk-one-login/account-interventions-service/blob/99284445cd91a3283de988a15bde497a1a900153/src/services/test/persist-intervention-events.test.ts#L82 and replace with an exact number

kirstyallen1:
> This all looks great.
> 
> Can we check for places like this
> 
> https://github.com/govuk-one-login/account-interventions-service/blob/99284445cd91a3283de988a15bde497a1a900153/src/services/test/persist-intervention-events.test.ts#L82
> 
> and replace with an exact number

Replaced all instances in unit tests with exact numbers in their assertions

edlouth (approved):

PR #1048, opened 2026-07-02
FPAD-8259: UI Interventions

edlouth:
## What

UI Interventions

## Notes

This PR is dependant on a chain of other PRs

### Issue tracking

- [FPAD-8259](https://govukverify.atlassian.net/browse/FPAD-8259)

[FPAD-8259]: https://govukverify.atlassian.net/browse/FPAD-8259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/frontend/views/index.njk
```diff
@@ -1,6 +1,21 @@
...
   <h1 class="govuk-heading-xl">Account Interventions Service</h1>
-  {{ govukButton({ text: "Get started" }) }}
+  <form method="post" action="{{ stagePrefix }}/search">
+    {{ govukInput({
+      label: {
+        text: "GOV.UK One Login Subject Identifier",
+        classes: "govuk-label--m"
+      },
+      hint: {
+        text: "For example, urn:fdc:gov.uk:2022:pgt1qOf7zW2tMCZg4V5LEu-mT-_GTSX7xqJ6RJekw9I"
```
ede-gds, line 13: **Question:** Where did this identifier come from?

src/frontend/app.ts
```diff
@@ -1,22 +1,40 @@
...
+const nodeModulesRoot = existsSync(path.join(__dirname, 'node_modules')) ? __dirname : process.cwd();
 
 // Stage prefix for asset URLs — empty string locally, /v1 when behind API Gateway without a custom domain
 const stagePrefix = process.env['STAGE_PREFIX'] ?? '';
 
-export function init() {
+const statusApiUrl = process.env['STATUS_API_URL'];
+
+const generateInterventionClient = () => {
+  if (!statusApiUrl) throw new Error('Missing required environment variable: STATUS_API_URL');
```
ede-gds: **Suggestion:** I wonder if we should leave it up to the `InterventionClient` to throw if it gets passed undefined

src/frontend/tests/app.test.ts
```diff
@@ -1,22 +1,103 @@
...
+        url: '/search',
+        payload: '',
+        headers: { 'content-type': 'application/x-www-form-urlencoded' },
+      });
+      expect(response.statusCode).toBe(303);
+      expect(response.headers.location).toBe('/user/');
+    });
+  });
+
+  describe('GET /user/:userId', () => {
```
ede-gds, line 63: **Suggestion:** Might be worth testing we get a 404 for not providing a user id / providing a non-existent one

src/frontend/app.ts
```diff
@@ -32,13 +50,38 @@ export function init() {
...
+/* istanbul ignore next */
+if (require.main === module) {
+  // called directly i.e. "node app"
+  init(
+    new InterventionStub({ interventionNames: [InterventionName.RESET_PASSWORD, InterventionName.REPROVE_IDENTITY] }),
+  ).listen({ port: 3000 }, (error) => {
+    if (error) console.error(error);
+    console.log('server listening on 3000');
+  });
+}
```
ede-gds: **Suggestion:** I think we got rid of this in a different PR?

```suggestion
```

src/frontend/app.ts
```diff
@@ -1,22 +1,40 @@
...
+import formbody from '@fastify/formbody';
 import nunjucks from 'nunjucks';
 import path from 'node:path';
 import { existsSync } from 'node:fs';
+import {
+  InterventionClient,
+  InterventionClientInterface,
+  InterventionName,
+  InterventionStub,
+} from '@govuk-one-login/ais-status-sdk';
```
ede-gds: **Thought:** Should we be retrieving these from where they live in the code? Slightly contrived example but if we need to support multiple versions of the API we'd need the ability to work with types from different versions of the SDK

ede-gds:
Looks good, just a few things

src/frontend/views/index.njk
edlouth: I have copied this from the FAI example, see https://github.com/govuk-one-login/fraud-admin-interface/blob/d1f9425c404f370f484e08d2e68f05396cdef3cc/src/frontend/views/add-fraud-intervention-page.njk#L37.

src/frontend/tests/app.test.ts
edlouth: I can add a test for `GET /user/` and check it returns a 404.

Currently if you put in any value for `:userId` you get a no active interventions response, this is how it is meant to work, given that the majority of requests are for accounts that have never had an intervention.
There is an open question about whether the status API should have a way of differentiating these from accounts where all the interventions have been cleared.

src/frontend/app.ts
edlouth: A good thought that probably applies for generally 👍

src/frontend/app.ts
edlouth: I am not sure I fully understand, as this is using the SDK I don't see why it would need to support multiple versions?  I don't think the SDK is going to let you call both v1 and v2 endpoints with one version of the SDK.

src/frontend/tests/app.test.ts
edlouth: Done

src/frontend/app.ts

src/frontend/app.ts
ede-gds: nvm I'm being silly

ede-gds (approved):

src/frontend/app.ts
```diff
@@ -56,7 +56,9 @@ export function init(
 
   // Fetches account status for the given userId and renders the details page.
   server.get<{ Params: { userId: string } }>('/user/:userId', async (request, reply) => {
-    const userId = decodeURIComponent(request.params.userId);
+    const userId = decodeURIComponent(request.params.userId).trim();
+
+    if (!userId) return reply.code(404).send();
```
ede-gds: sorry I approved but thinking about it this should be a 400

```6.5.1. 400 Bad Request
The 400 (Bad Request) status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).```

PR #1052, opened 2026-07-02
Allow npm published to bypass branch protection

ede-gds:
## What

Allows the new npm package publishing Github app to bypass `main` branch protection rules

## Why

We want to automatically update the main branch when we publish.

### DIFF
--- a/.github/rulesets/protect_main.json
+++ b/.github/rulesets/protect_main.json
@@ -62,5 +62,13 @@
         ]
       }
     }
-  ]
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 4175315,
+      "actor_type": "Integration",
+      "bypass_mode": "always"
+    }
+  ],
+  "current_user_can_bypass": "never"
 }


.github/rulesets/protect_main.json
```diff
@@ -62,5 +62,13 @@
         ]
       }
     }
-  ]
+  ],
+  "bypass_actors": [
+    {
+      "actor_id": 4175315,
```
edlouth, line 68: I presume this is correct, but it doesn't match the bot id

edlouth (approved):

.github/rulesets/protect_main.json
ede-gds: hm, interesting. might be worth trying to find out where this comes from before we merge

ede-gds: this is still on evaluate anyway so I don't mind delaying a touch if you don't

edlouth: I have checked, this is the App ID, so all good 👍

ede-gds: nice, thanks

PR #1056, opened 2026-07-03
FPAD-8429: Improvements to update command

edlouth:
## What

Move DynamoDB aspects to dynamo-db-record-service

## Why

Collecting the DynamoDB aspects in one place makes it easier to swap out different components.

As changes to DynamoDB queries is higher risk, we want to reduce these changes and in future we can have test coverage for these methods against a proper DynamoDB mock.

## Notes

If `ConditionExpression` contains any variables these need to be manually set with `ExpressionAttributeValues`.

### Issue tracking

- [FPAD-8429](https://govukverify.atlassian.net/browse/FPAD-8429)

[FPAD-8429]: https://govukverify.atlassian.net/browse/FPAD-8429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/db-helpers.ts
+++ b/src/services/db-helpers.ts
@@ -1,0 +1,13 @@
+export const generateExpressionAttributeNames = (keys: string[]) =>
+  Object.fromEntries(keys.map((key) => [`#${key}`, key]));
+
+export const generateExpressionAttributeValues = (input: Record<string, unknown>) =>
+  Object.fromEntries(Object.entries(input).map(([key, value]) => [`:${key}`, value]));
+
+export const generateUpdateExpressionClause = (keys: string[], keysToRemove?: string[]) =>
+  `SET ${keys.map((key) => keyToUpdateExpressionClause(key)).join(', ')}${keysToRemoveExpressionClause(keysToRemove)}`;
+
+const keyToUpdateExpressionClause = (key: string) => `#${key} = :${key}`;
+
+export const keysToRemoveExpressionClause = (keysToRemove?: string[]) =>
+  keysToRemove && keysToRemove.length > 0 ? ` REMOVE ${keysToRemove.join(', ')}` : '';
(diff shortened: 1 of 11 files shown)


src/services/test/db-helpers.test.ts
```diff
@@ -0,0 +1,19 @@
+import { keysToRemoveExpressionClause } from '../db-helpers';
+
+describe('keysToRemoveExpressionClause', () => {
```
kirstyallen1, line 3: Question: Should we be testing the other db helper functions directly? I assume they're probably covered indirectly in other tests as well but just wondering

src/services/db-helpers.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const generateExpressionAttributeValues = (input: Record<string, unknown>) =>
+  Object.fromEntries(Object.entries(input).map(([key, value]) => [`:${key}`, value]));
+
+export const generateUpdateExpressionClause = (keys: string[], keysToRemove?: string[]) =>
+  `SET ${keys.map((key) => keyToUpdateExpressionClause(key)).join(', ')}${keysToRemoveExpressionClause(keysToRemove)}`;
+
+const keyToUpdateExpressionClause = (key: string) => `#${key} = :${key}`;
+
+export const keysToRemoveExpressionClause = (keysToRemove?: string[]) =>
```
kirstyallen1, line 12: Question: Is this okay to just use the bare name, or should it use aliases similar to how the `keyToUpdateExpressionClause` does?

edlouth: Well this is interesting, turns out both `SET` and `REMOVE` only need to use `ExpressionAttributeNames` when they are reserved words. Things work as they are now, however you might be right that we should also use `ExpressionAttributeNames` for remove, which should cover all cases.

src/services/test/db-helpers.test.ts
edlouth: So we do have 100% coverage for this change, potentially I should also run it through mutation testing.  However I think I will leave this for another day, if someone is keen.

kirstyallen1 (approved):

PR #1072, opened 2026-07-08
FPAD-8480: Refactor of Message Service

edlouth:
## What

Refactor of Message Service

## Why

Follow ports and adaptors model.

### Issue tracking

- [FPAD-8480](https://govukverify.atlassian.net/browse/FPAD-8480)

[FPAD-8480]: https://govukverify.atlassian.net/browse/FPAD-8480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/handlers/txma-handler.ts
```diff
@@ -1,11 +1,15 @@
...
 import { addMetric, metric } from '../commons/metrics';
 import { MetricNames } from '../data-types/constants';
 import jsonSafeParse from '../commons/json-safe-parse';
 
-export async function handler(event: SQSEvent, context: Context): Promise<void> {
+export async function handler(
+  event: SQSEvent,
+  context: Context,
+  messageService: MessageServiceInterface = new MessageService(),
+): Promise<void> {
```
ede-gds: @kirstyallen1 just a heads up on this change in case this branch gets in before you finish FPAD-8430

src/services/test/message-service.test.ts
```diff
@@ -20,42 +20,48 @@ vi.mock('@aws-sdk/client-sqs', async (importOriginal) => {
 });
 vi.mock('../../commons/logger');
 
+describe('MessageService', () => {
+  it('throws an error if AWS_REGION environment variable is not set', () => {
+    vi.stubEnv('AWS_REGION', '');
+
+    expect(() => new MessageService()).toThrow('Environment variable AWS_REGION not found');
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    expect(vi.mocked(logger.error)).toHaveBeenCalledWith('Environment variable AWS_REGION not found');
```
ede-gds: **Suggestion:** I know this is from elsewhere but I think we can get rid of these lines

```suggestion
```

src/services/message-service.ts
```diff
@@ -0,0 +1,65 @@
...
+import tracer from '../commons/tracer';
+import getEnvironmentOrThrow from '../commons/get-environment-or-throw';
+
+export function createSqsClient() {
+  const region = getEnvironmentOrThrow('AWS_REGION');
+
+  return tracer.captureAWSv3Client(new SQSClient({ region }));
+}
+
+export interface MessageServiceInterface {
```
ede-gds: **Suggestion (non-blocking):** This is personal preference but I would call the interface `MessageService` then have classes called `SQSMessageService` and `StubMessageService` or `InMemoryMessageService`

src/services/message-service.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export class MessageStub implements MessageServiceInterface {
+  constructor(
+    private readonly messageOutput?: SendMessageCommandOutput,
+    private readonly batchMessageOutput?: SendMessageBatchCommandOutput,
+  ) {}
+
+  sendMessage() {
+    if (!this.messageOutput) return Promise.reject(new Error('Send message error'));
```
ede-gds: **Suggestion:** Make this error more obvious in case someone messes up their test setup:

```suggestion
    if (!this.messageOutput) return Promise.reject(new Error('No message output provided to stub'));
```

src/services/message-service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    return Promise.resolve(this.messageOutput);
+  }
+
+  sendBatchMessage() {
+    if (!this.batchMessageOutput) return Promise.reject(new Error('Send batch message error'));
```
ede-gds: **Suggestion:** Similarly

```suggestion
    if (!this.batchMessageOutput) return Promise.reject(new Error('No batch message output provided to stub'));
```

src/services/message-service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class MessageService implements MessageServiceInterface {
+  constructor(
+    readonly client: SQSClient = createSqsClient(),
+    readonly queueUrl?: string,
+  ) {}
+
+  sendMessage(messageBody: object | string, queueUrl: string | undefined) {
+    const message: SendMessageRequest = {
+      QueueUrl: queueUrl ?? this.queueUrl,
```
ede-gds: **Suggestion:** I think it's easier to understand and test if a given message client has a single `queueUrl` and this function and `sendBatchMessage` have a single `messageBody` argument.

ede-gds (commented):
Lovely stuff

src/services/test/message-service.test.ts
edlouth: Done

src/services/message-service.ts
edlouth: Agreed

src/services/message-service.ts
edlouth: Currently this is called with two different queues https://github.com/govuk-one-login/account-interventions-service/pull/1072/changes#diff-9fc8b4553f9e65912a3ef219c85edb8fd63beece0fd564d22525b10553b02eb8R70

ede-gds: I think we should instantiate two separate objects - the alternative is going functional as before (and pass in the functions so we can test) but in this case I reckon it's nicer to encapsulate and just pass around two objects

ede-gds (approved):
lgtm 🎉

PR #1073, opened 2026-07-08
FPAD-8470: Frontend messages

edlouth:
## What

Frontend messages

## Why

Preparing for interventions update UI

## Notes

Access to send TxMA messages only applies as far as `build` as part of this PR.

### Issue tracking

- [FPAD-8470](https://govukverify.atlassian.net/browse/FPAD-8470)

[FPAD-8470]: https://govukverify.atlassian.net/browse/FPAD-8470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/frontend/app.ts
```diff
@@ -62,9 +73,50 @@ export function init(
...
+
+    const event: TicfAccountIntervention = {
+      event_id: randomUUID(),
+      event_name: TriggerEventsEnum.TICF_ACCOUNT_INTERVENTION,
+      component_id: 'AIS',
+      timestamp: timestamp.seconds,
+      event_timestamp_ms: timestamp.milliseconds,
+      user: { user_id: userId },
+      extensions: {
+        intervention: {
```
kirstyallen1, line 132: Could we remove the hardcoded values in here?

edlouth: I have removed my email, otherwise this is just a simple demo of the SQS functionality.  We will replace this later with a more fully formed interface for raising interventions.

samconfig.toml
```diff
@@ -22,3 +22,16 @@ image_repositories = []
...
+resolve_s3 = true
+s3_prefix = "ais-frontend-eatl2"
+region = "eu-west-2"
+confirm_changeset = true
+capabilities = "CAPABILITY_NAMED_IAM"
+parameter_overrides = "Environment=\"dev\" ApiStageName=\"v1\" StagePrefix=\"/v1\" VpcStackName=\"vpc\" MainStackName=\"ais-main\" PermissionsBoundary=\"none\" CSLSSubscriptionEndpointArn=\"arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython-2\" CodeSigningConfigArn=\"none\" ProductTagValue=\"GOV.UK One Login\" SystemTagValue=\"Account Interventions Service\" OwnerTagValue=\"interventions@digital.cabinet-office.gov.uk\" SourceTagValue=\"govuk-one-login/account-interventions-service/src/infra/main/template.yaml\""
+image_repositories = []
+
+[frontend-eatl2.global.parameters]
+region = "eu-west-2"
```
ede-gds: **Suggestion:** Can we do one of two things:

- have the `eatl-2` configs live on a stash on your machine
- call this something descriptive (`frontend-scratch`, `frontend-alpha`)

It would also be great if we could put changes like this in a proper commit so future team members know what's going on - `Basic event send button` isn't very helpful for actually understanding why this config exists.

src/frontend/app.ts
```diff
@@ -62,9 +73,50 @@ export function init(
...
+    }
+
     const accountStatus = await interventionClient.getAccountStatus(userId);
 
-    return reply.view('user-details.njk', { stagePrefix, assetPath, accountStatus, userId });
+    return reply.view('user-details.njk', { stagePrefix, assetPath, accountStatus, userId, messageSent });
+  });
+
+  // Sends a TxMA audit event for the given userId and redirects back to the user details page.
+  server.post<{ Body: { userId?: string } }>('/send', async (request, reply) => {
```
ede-gds, line 117: **Suggestion:** We definitely don't want this in production (as it is here) so this should be behind a feature flag by default (i.e. if there's no flag set it should be disabled) and throw a 404 if it is disabled

src/frontend/views/user-details.njk
```diff
@@ -25,4 +33,8 @@
   {% else %}
     <p class="govuk-body">No account found for this identifier.</p>
   {% endif %}
-{% endblock %}
+  <form method="post" action="{{ stagePrefix }}/send">
+    <input type="hidden" name="userId" value="{{ userId }}">
+      {{ govukButton({ text: "Send TxMA Message" }) }}
+    </form>
```
ede-gds, line 85: **Suggestion:** This form should also be conditional on a feature flag and hidden by default

src/frontend/app.ts
```diff
@@ -16,18 +22,23 @@ const nodeModulesRoot = existsSync(path.join(__dirname, 'node_modules')) ? __dir
 const stagePrefix = process.env['STAGE_PREFIX'] ?? '';
 
 const statusApiUrl = process.env['STATUS_API_URL'];
+const txmaQueueUrl = process.env['TXMA_QUEUE_URL'];
```
ede-gds, line 26: **Suggestion (non-blocking):** Is it worth using `getEnvOrThrow` for some of / any of these?

src/frontend/app.ts
```diff
@@ -62,9 +73,50 @@ export function init(
 
     if (!userId) return reply.code(400).send();
 
+    // Read and immediately clear the flash cookie so the banner only shows once.
+    // eslint-disable-next-line unicorn/consistent-boolean-name
+    const messageSent = request.cookies['flash_message_sent'] === 'true';
+    if (messageSent) {
+      void reply.clearCookie('flash_message_sent', { path: '/' });
+    }
```
ede-gds, line 97: **Thought:** This feels a bit clumsy - is this from prior art? Can we not get away with a query param or something?

ede-gds (commented):
Good stuff, a few thoughts and comments

samconfig.toml
edlouth: I will remove this for now, I am not sure I entirely intended to commit it.

In reality this is really a deployment just for this PR.

src/frontend/app.ts
edlouth: Both of these are handled in services, so there are no checks in this file, and indeed they are not set when run locally.

Potentially this would be resolved by separating out a Lambda handler as @kirstyallen1 has done for the other Lambda functions.

For now I will leave as they are.

src/frontend/app.ts
edlouth: Open to suggestions.

This is a fairly common pattern when you want a banner to show only once, a query param would work, but then when you refresh the page the banner stays.  I could potentially look to abstract this in some way.  I think this is roughly how FAI works at times.

src/frontend/views/user-details.njk
edlouth: Done

src/frontend/app.ts
edlouth: Done

src/infra/frontend/template.yaml
```diff
@@ -176,6 +182,12 @@ Resources:
             - PrivateApiId:
                 Fn::ImportValue: !Sub '${MainStackName}-PrivateApiRestApiId'
           ENABLE_AIS_FRONTEND: !FindInMap [FeatureFlags, !Ref Environment, EnableAISFrontend]
+          ENABLE_AIS_SEND_TXMA: !FindInMap [FeatureFlags, !Ref Environment, EnableAISSendTxMA]
+          TXMA_QUEUE_URL: !If
+            - IsInternal
+            - Fn::ImportValue: !Sub '${MainStackName}-TxMAIngressSqsQueueUrl'
+            - !Ref 'AWS::NoValue'
+            # - Fn::ImportValue: !Sub '${MainStackName}-TxMAEgressSqsQueueUrl'
```
ede-gds, line 190: **Question:** are these comments actually doing anything?

ede-gds (approved):
lgtm, thanks!

src/infra/frontend/template.yaml
edlouth: No, they are just there for when we want this to work in higher environments.

PR #1095, opened 2026-07-15
FPAD-8468: Allow FAI to access AIS frontend

edlouth:
## What

Allow FAI to access AIS frontend

Handle `STAGE_PREFIX` and `SUBPATH` for different Public API Gateway setups

### Issue tracking

- [FPAD-8468](https://govukverify.atlassian.net/browse/FPAD-8468)

[FPAD-8468]: https://govukverify.atlassian.net/browse/FPAD-8468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commons/test/normalise-path-segment.test.ts
+++ b/src/commons/test/normalise-path-segment.test.ts
@@ -1,0 +1,28 @@
+import { normalisePathSegment } from '../utils/normalise-path-segment';
+
+describe('normalisePathSegment', () => {
+  it.each([
+    // empty / whitespace
+    ['', ''],
+    [' '.repeat(3), ''],
+
+    // no leading slash — one is prepended
+    ['v1', '/v1'],
+    ['interventions/v1', '/interventions/v1'],
+
+    // leading slash already present — preserved as-is
+    ['/v1', '/v1'],
+    ['//v1', '//v1'],
+
+    // trailing slash stripped (only the last character)
+    ['/v1/', '/v1'],
+    ['interventions/', '/interventions'],
+    ['//v1//', '//v1/'],
+
+    // whitespace trimmed before normalising
+    ['  /v1  ', '/v1'],
+    ['  interventions/  ', '/interventions'],
+  ])('normalisePathSegment(%j) → %j', (input, expected) => {
+    expect(normalisePathSegment(input)).toBe(expected);
+  });
+});
--- a/src/commons/utils/normalise-path-segment.ts
+++ b/src/commons/utils/normalise-path-segment.ts
@@ -12,7 +12,7 @@
  */
 export function normalisePathSegment(value: string): string {
   let trimmed = value.trim();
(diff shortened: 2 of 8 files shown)


src/frontend/app.ts
```diff
@@ -73,14 +83,19 @@ export function init(
     templates: [path.join(__dirname, 'views'), path.join(nodeModulesRoot, 'node_modules/govuk-frontend/dist')],
   });
 
-  const assetPath = `${stagePrefix}/assets`;
+  /**
+   * urlPrefix combines subpath and stagePrefix for constructing page-level navigation URLs.
+   * Templates and redirects use this so links resolve correctly through the API Gateway subpath.
+   */
+  const urlPrefix = `${subpath}${stagePrefix}`;
```
ede-gds: **Nitpick:** quite pedantic but `urlPrefix` implies to me that it's on the front of a url with a domain etc - I think `pathPrefix` makes it more obvious to me what this is?

src/commons/utils/normalise-path-segment.ts
```diff
@@ -0,0 +1,18 @@
...
+ *   "/v1"          → "/v1"
+ *   "/v1/"         → "/v1"
+ *   "interventions/" → "/interventions"
+ */
+export function normalisePathSegment(value: string): string {
+  let trimmed = value.trim();
+  if (trimmed.endsWith('/')) trimmed = trimmed.slice(0, -1); // strip trailing slashes
+  if (!trimmed) return '';
+  return trimmed.startsWith('/') ? trimmed : `/${trimmed}`;
+}
```
ede-gds, line 18: **Suggestion:** This looks perfectly sensible but you might be better off using `Path`: https://nodejs.org/api/path.html

src/commons/test/normalise-path-segment.test.ts
```diff
@@ -0,0 +1,76 @@
+import { normalisePathSegment } from '../utils/normalise-path-segment';
+
+describe('normalisePathSegment', () => {
```
ede-gds, line 3: **Suggestion:** You could probably parameterise this whole suite into a single test (quite suprised that sonar isn't complaining about this)

ede-gds (approved):
lgtm, nothing blocking but I'd give `Path` a go because I remember it being actually very good

src/frontend/app.ts

src/commons/utils/normalise-path-segment.ts
edlouth: The suggestion is not to use `path` for this, as `path` is designed for file paths rather than URLs

src/commons/test/normalise-path-segment.test.ts
edlouth: Done

PR #1097, opened 2026-07-17
FPAD-8463: AIS UI JWT auth

edlouth:
## What

AIS UI JWT auth

## Why

Zero trust architecture

## Notes

Need to pass in `FaiAuthSigningKeyArn` to frontend CloudFormation template.

### Issue tracking

- [FPAD-8463](https://govukverify.atlassian.net/browse/FPAD-8463)

[FPAD-8463]: https://govukverify.atlassian.net/browse/FPAD-8463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edlouth:
Potentially `FaiAuthSigningKeyArn` should move from a parameter to a mapping. Just means the values are set inside [frontend/template.yaml](https://github.com/govuk-one-login/account-interventions-service/blob/main/src/infra/frontend/template.yaml) here rather than in [ais-base-infrastructure](https://github.com/govuk-one-login/ais-base-infrastructure).

src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,95 @@
+import { KMSClient, GetPublicKeyCommand } from '@aws-sdk/client-kms';
+import { importSPKI, jwtVerify, type JWTPayload } from 'jose';
+import logger from '../commons/logger';
+import getEnvironmentOrThrow from '../commons/get-environment-or-throw';
+
+export enum Role {
+  STANDARD_USER = 'standard-user',
+  ADMINISTRATOR = 'administrator',
+  JOINT_CSV_UPLOAD = 'joint-csv-upload',
+}
```
ede-gds, line 11: **Question:** Are we going to have to keep this up to date with FAI?

src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Converts a DER-encoded SubjectPublicKeyInfo buffer to a PEM string.
+ */
+function derToPem(der: Uint8Array): string {
+  const b64 = Buffer.from(der).toString('base64');
+  const lines = b64.match(/.{1,64}/g) ?? [];
+  return `-----BEGIN PUBLIC KEY-----\n${lines.join('\n')}\n-----END PUBLIC KEY-----\n`;
+}
```
ede-gds, line 101: **Suggestion:** Seems like we can do this with the standard library: https://stackoverflow.com/a/76291232

src/frontend/authoriser.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface Authoriser {
+  verify(request: FastifyRequest): Promise<AuthoriserResult>;
+}
+
+export class JwtAuthoriser implements Authoriser {
+  constructor(readonly jwtVerifier: JwtVerifierInterface = new KmsJwtVerifier()) {}
+
+  async verify(request: FastifyRequest): Promise<AuthoriserResult> {
+    const authorizer = request.awsLambda.event.requestContext.authorizer as
+      { jwt?: string; principalId?: string } | undefined;
```
ede-gds: **Suggestion (non-blocking):** Make a little zod type to parse this

src/frontend/authoriser.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+export interface Authoriser {
+  verify(request: FastifyRequest): Promise<AuthoriserResult>;
+}
+
+export class JwtAuthoriser implements Authoriser {
+  constructor(readonly jwtVerifier: JwtVerifierInterface = new KmsJwtVerifier()) {}
+
+  async verify(request: FastifyRequest): Promise<AuthoriserResult> {
```
ede-gds: **Suggestion (non-blocking):** This class is quite coupled to fastify - it might help to abstract this away and just have `verify` take whatever `request.awsLambda.event.requestContext.authorizer` is (see other comment about parsing with Zod)

src/frontend/authoriser.ts
```diff
@@ -0,0 +1,51 @@
...
+    const token = authorizer?.jwt;
+
+    if (!token) {
+      logger.warn('Request has no JWT in authorizer context', { url: request.url });
+      return { success: false };
+    }
+
+    try {
+      const payload = await this.jwtVerifier.verify(token);
+      logger.info('JWT verified successfully', { sub: payload.sub, url: request.url });
```
ede-gds: **Issue:** Probably shouldn't be logging email addresses here

src/frontend/app.ts
```diff
@@ -59,6 +67,7 @@ export function init(
   }),
   featureFlags: FeatureFlags = FeatureFlagsFromEnvironmentVariables.getInstance(),
   messageService: MessageService = new SqsMessageService(txmaQueueUrl),
+  authoriser: Authoriser = new JwtAuthoriser(),
```
ede-gds, line 77: **Suggestion (non-blocking):** Would be good to move away from this pattern of defaults, and instead just construct the production dependencies at the top level of the Lambda handler. Not necessarily for this PR though

src/frontend/app.ts
```diff
@@ -144,17 +163,26 @@ export function init(
...
+        error: 'Missing interventionCode',
+        message: 'An intervention code is required. Please select an intervention code before submitting.',
+      });
     }
 
     if (!isCode(interventionCode)) {
-      return reply.code(422).send({ error: 'Invalid interventionCode', message: `"${interventionCode}" is not a recognised intervention code. Please select a valid code from the list.` });
+      return reply.code(422).send({
+        error: 'Invalid interventionCode',
+        message: `"${interventionCode}" is not a recognised intervention code. Please select a valid code from the list.`,
```
ede-gds: **Nitpick:** This doesn't own the frontend representation so we shouldn't refer to a list

```suggestion
        message: `"${interventionCode}" is not a recognised intervention code. Please provide a valid code.`,
```

src/frontend/tests/authoriser.test.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// JwtAuthoriser
+// ---------------------------------------------------------------------------
+
+describe('JwtAuthoriser', () => {
+  let verifyMock: ReturnType<typeof vi.fn<(token: string) => Promise<FaiJwtPayload>>>;
+  let stubVerifier: JwtVerifierInterface;
+  let authoriser: JwtAuthoriser;
```
ede-gds: **Suggestion (non-blocking):** Like we spoke about, maybe get rid of these `let`s in favour of initialising things inside the test block

src/services/jwt-verifier.ts
edlouth: Yes I think so, however I think these are fairly slow changing

src/services/jwt-verifier.ts
edlouth: Indeed, updated

src/frontend/authoriser.ts
edlouth: Fixed

src/frontend/authoriser.ts
edlouth: Fixed

src/frontend/authoriser.ts
edlouth: Done

src/frontend/app.ts

src/frontend/tests/authoriser.test.ts
edlouth: Done

src/frontend/authoriser.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export class JwtAuthoriser implements Authoriser {
+  constructor(readonly jwtVerifier: JwtVerifierInterface = new KmsJwtVerifier()) {}
+
+  async verify(authoriserContext: AuthoriserContext, url: string): Promise<AuthoriserResult> {
+    const context = this.validateContext(authoriserContext);
+
+    if (!context.success) {
+      logger.warn('Request has no JWT in authorizer context', { url });
```
ede-gds: **Nitpick:** Not necessarily no JWT here (could be the wrong type, etc)

```suggestion
      logger.warn('Request has invalid or missing JWT in authorizer context', { url });
```

src/services/jwt-verifier.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Converts a DER-encoded SubjectPublicKeyInfo buffer to a PEM string.
+ */
+function derToPem(der: Uint8Array): string {
+  const key = createPublicKey({ key: Buffer.from(der), format: 'der', type: 'spki' });
+
+  const output = key.export({ format: 'pem', type: 'spki' });
+
+  if (typeof output !== 'string')
+    throw new Error('Error converting  DER-encoded SubjectPublicKeyInfo buffer to a PEM string');
```
ede-gds: **Nitpick:**
```suggestion
    throw new Error('Error converting DER-encoded SubjectPublicKeyInfo buffer to a PEM string');
```

ede-gds (approved):
lgtm, just two strings

src/frontend/authoriser.ts
edlouth: Done

src/services/jwt-verifier.ts
edlouth: Fixed

PR #1109, opened 2026-07-20
FPAD-8545: Update form to raise different interventions

edlouth:
## What

Update form to raise different interventions

## Why

Will make it easier to setup accounts in dev for demos

## Notes

To follow with more fields potentially

### Issue tracking

- [FPAD-8545](https://govukverify.atlassian.net/browse/FPAD-8545)

[FPAD-8545]: https://govukverify.atlassian.net/browse/FPAD-8545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/message-service.ts
```diff
@@ -64,9 +64,12 @@ export class StubMessageService implements MessageService {
     private readonly batchMessageOutput?: SendMessageBatchCommandOutput,
   ) {}
 
-  sendMessage() {
+  sendMessage(messageBody: object) {
     if (!this.messageOutput) return Promise.reject(new Error('No message output provided to stub'));
 
+    console.log('Sending message', { messageBody });
```
kirstyallen1: Is this console log needed?

edlouth: A good question. I put this in as this is a stub and it was useful at the time, however it probably logs out a whole load of stuff during tests, which we don't really need.  I think I will remove it, mainly because it is always easy to put this kind of thing back in.

src/frontend/app.ts
```diff
@@ -126,14 +134,23 @@ export function init(
...
-    const userId = request.body.userId?.trim() ?? '';
+    const userId = request.body.userId?.trim();
+
+    if (!userId) throw new Error('Missing userId');
+
+    const interventionCode = request.body.interventionCode;
+
+    if (!interventionCode) throw new Error('Missing interventionCode');
+
+    if (!isCode(interventionCode)) throw new Error('Invalid interventionCode');
```
kirstyallen1: Question: I know this is just an internal tool for dev etc. but would rendering something with these error codes instead of just having a stack trace showing be useful?

edlouth: Good thought, I should return a proper 422 response 👍

src/frontend/views/user/_active-interventions.njk
```diff
@@ -0,0 +1,18 @@
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+<h1 class="govuk-heading-l">Active Interventions</h1>
```
kirstyallen1: Suggestion: This should maybe be a `h2` tag. If this template is imported into `user-details.njk` then there will be two `h1` elements on that page I think?

src/frontend/views/user/_send-txma-form.njk
```diff
@@ -0,0 +1,16 @@
...
+{% if aisSendTxMA %}
+  <h2 class="govuk-heading-m">Send Intervention</h2>
+  <form method="post" action="{{ stagePrefix }}/send">
+    <input type="hidden" name="userId" value="{{ userId }}"/> {{ govukSelect({
+      id: "intervention-code",
+      name: "interventionCode",
+      label: {
+        text: "Intervention type"
+      },
+      items: interventions
```
kirstyallen1, line 12: Thought: Perhaps not necessary for the current use case but maybe a default placeholder in this dropdown could be good

edlouth: Currently the first option shows to start with.  I don't think there is a need to go to the length of creating a placeholder, so that a user has to actively choose.  This dropdown isn't the long term solution for this UX.

src/frontend/views/user/_active-interventions.njk
edlouth: Done

src/frontend/app.ts

kirstyallen1 (approved):

PR #1116, opened 2026-07-22
FPAD-8497: Improve testability of business logic

kirstyallen1:
## What
We want to improve the testability of our business logic that drives our lambdas by doing a few things:
- Ensure the handlers only make a single function call.
- Business logic functions take all their pre-requisites as function args.
- Business logic functions are tested robustly using injected fakes instead of runtime patching.

## Why
With these changes we can test our business logic independently and without needing any runtime patching. We aim to test the lambda handlers themselves in the feature tests rather than via unit tests.

### Issue tracking

- [FPAD-8497](https://govukverify.atlassian.net/browse/FPAD-8497)

[FPAD-8497]: https://govukverify.atlassian.net/browse/FPAD-8497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-project.properties
```diff
@@ -9,4 +9,4 @@ sonar.tests=src/
 sonar.test.inclusions=**/*.test.ts
 sonar.sourceEncoding=UTF-8
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
-sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts, src/frontend/local.ts
+sonar.coverage.exclusions=src/handlers/invoke-private-api.ts, src/contract-testing/**, src/data-types/api-schemas-v*.ts, src/frontend/local.ts, src/handlers/interventions-processor-handler.ts, src/handlers/txma-handler.ts
```
kirstyallen1, line 12: Sonar is flagging these two handlers due to some code changes, even with the pragmas in place. Is there another solution to ensure sonar is not flagging these files or is this the best way?

src/handlers/tests/interventions-processor-handler.test.ts
```diff
@@ -106,7 +102,7 @@ describe('intervention processor handler', () => {
       awsRegion: '',
     };
     mockEvent = { Records: [mockRecord] };
-    accountStateEngine.getInterventionEnumFromCode = vi.fn().mockImplementation(() => EventsEnum.FRAUD_BLOCK_ACCOUNT);
+    // accountStateEngine.getInterventionEnumFromCode = vi.fn().mockImplementation(() => EventsEnum.FRAUD_BLOCK_ACCOUNT);
```
edlouth: I think lets remove this

```suggestion
```

src/handlers/tests/interventions-processor.test.ts
```diff
@@ -247,24 +290,21 @@ describe('intervention processor handler', () => {
...
+            accountStateEngine,
+          }
+        ),
       ).toEqual({
         batchItemFailures: [],
       });
       expect(sendAuditEvent).toHaveBeenLastCalledWith(
         'AIS_EVENT_TRANSITION_APPLIED',
-        EventsEnum.FRAUD_BLOCK_ACCOUNT,
+        EventsEnum.FRAUD_SUSPEND_ACCOUNT,
```
edlouth, line 306: Is there a good reason to change these tests?

src/handlers/tests/interventions-processor-handler.test.ts
kirstyallen1: Good spot - will remove

src/handlers/tests/interventions-processor.test.ts
kirstyallen1: The tests previously used a mocked return value from the accountStateEngine. Once I removed the mock I found that the input going into the tests didn't align with the assertions but this didn't have any effect due to the mocked return. These tests have just been adjusted now that the real state engine is being used so the assertions have mostly changed rather than the input

edlouth: Sounds fine to me 👍

edlouth (approved):
If you can try the v8 ignore, but otherwise all looks good to me 👍

ede-gds:
Looks good to me but just one small thing I noticed - the file name of `src/handlers/tests/interventions-processor-handler.test.ts` should be updated to `src/handlers/tests/interventions-processor.test.ts`

kirstyallen1:
> Looks good to me but just one small thing I noticed - the file name of `src/handlers/tests/interventions-processor-handler.test.ts` should be updated to `src/handlers/tests/interventions-processor.test.ts`

Good spot - I'll update all these file names to match

PR #1119, opened 2026-07-22
FPAD-8586: Exclude ais-status-sdk integration-tests folder from coverage report

kirstyallen1:
## What
Exclude the contents of `packages/ais-status-sdk/integration-tests/` from coverage reports when `npm test` is run. The `test-server.ts` file should be excluded as this is simply a test harness, rather than code that needs testing.

## Why
We only want to show relevant files in the coverage report

### Issue tracking

- [FPAD-8586](https://govukverify.atlassian.net/browse/FPAD-8586)

[FPAD-8586]: https://govukverify.atlassian.net/browse/FPAD-8586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -15,7 +15,7 @@ export default defineConfig({
       reporter: ['text', 'lcov'],
       reportsDirectory: './coverage',
     },
-    exclude: ['**/node_modules/**', 'src/contract-testing/**', '.stryker-tmp/**', 'src/scripts/**'],
+    exclude: ['**/node_modules/**', 'src/contract-testing/**', '.stryker-tmp/**', 'src/scripts/**', 'packages/ais-status-sdk/integration-tests/**'],
     env: {
       CLOUDWATCH_METRICS_NAMESPACE: 'test_namespace',
       METRIC_SERVICE_NAME: 'test',


vitest.config.ts
```diff
@@ -14,6 +14,7 @@ export default defineConfig({
       provider: 'v8',
       reporter: ['text', 'lcov'],
       reportsDirectory: './coverage',
+      exclude: ['packages/ais-status-sdk/integration-tests/**'],
     },
     exclude: ['**/node_modules/**', 'src/contract-testing/**', '.stryker-tmp/**', 'src/scripts/**'],
```
edlouth: I think we can move this into the existing `exclude` bellow.

```suggestion
},
exclude: ['**/node_modules/**', 'src/contract-testing/**', '.stryker-tmp/**', 'src/scripts/**', 'packages/ais-status-sdk/integration-tests/**'],
```

kirstyallen1: Changed :) I could have sworn this didn't work when I added it to this exclude y'day but seems to be working fine now so have amended!

edlouth (approved):

PR #1120, opened 2026-07-22
FPAD-8587: Extend app config class to return static config

ede-gds:
## What

Extends the app config class to create a static config object. The config object only contains config which you ask for, and the output is narrowly typed

## Why

The underlying config variables on a Lambda don't change after it's published ([see the docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html)) so there's no point in us referring to and validating config on every read. Instead I've extended the app-config class to let us build a statically typed config object at start time, checking and validating all the types then. We can deprecate the individual function calls for config variables as we start to roll this out on the handlers.

This has a few other benefits:

- Handlers and the functions they call can take static config objects so we don't need environment variables for testing, aside from for testing this stuff.
- We can only build the config with the values we need, so the config for a given handler is "safe".
- The config schema is totally data-driven, so to add a new config variable you don't need to add any code (assuming the type of the variable you're adding is already handled).

### Issue tracking

- [FPAD-8587](https://govukverify.atlassian.net/browse/FPAD-8587)

[FPAD-8587]: https://govukverify.atlassian.net/browse/FPAD-8587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/app-config-service.ts
+++ b/src/services/app-config-service.ts
@@ -152,11 +152,12 @@ export class AppConfigService {
     return this.instance;
   }
 
-  public get metricServiceName(): string | undefined {
+  public get metricServiceName(): string {
     try {
       return this.validateConfiguration('METRIC_SERVICE_NAME');
     } catch {
       logger.warn('Unable to retrieve metrics config, METRIC_SERVICE_NAME not set');
+      return 'unset';
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/services/app-config-service.ts
```diff
@@ -20,7 +153,12 @@ export class AppConfigService {
   }
 
   public get metricServiceName(): string {
-    return this.validateConfiguration('METRIC_SERVICE_NAME');
+    try {
+      return this.validateConfiguration('METRIC_SERVICE_NAME');
+    } catch {
+      logger.warn('Unable to retrieve metrics config, METRIC_SERVICE_NAME not set');
+      return 'unset';
```
edlouth, line 160: This makes me nervous, could we not return `undefined`?

edlouth (approved):
Overall looks good, just one question

src/services/app-config-service.ts
ede-gds: Returning `undefined` makes sonar complain (something about getters having to return a value) - when I was returning `undefined` this is actually what the metric service was doing in the background. We should probably hook this log up to something but other than that it doesn't make me particularly anxious - what were you thinking could go wrong?

edlouth: It is just using a constant as a text string, not a blocker.

PR #1138, opened 2026-07-27
FPAD-8592: Consistently display fields for each intervention

kirstyallen1:
## What
- Optional fields are now always shown with 'N/A' if there is no data to display. 
- There is now a field description section detailing meanings for originatingComponent, originatorReferenceId, and requesterId. 
- Sent at heading has been removed and is now rendered without a field name

## Why
This has changed as a result of user feedback and suggested improvements 

### Issue tracking

- [FPAD-8592](https://govukverify.atlassian.net/browse/FPAD-8592)

[FPAD-8592]: https://govukverify.atlassian.net/browse/FPAD-8592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/frontend/views/user/_history.njk
+++ b/src/frontend/views/user/_history.njk
@@ -3,25 +3,17 @@
 {% if accountHistory and accountHistory.length %}
   {% for transaction in accountHistory %}
     <div class="govuk-inset-text govuk-!-margin-bottom-5">
-      <p class="govuk-body-xs govuk-hint govuk-!-margin-bottom-4">Sent at:
+      <p class="govuk-body-xs govuk-hint govuk-!-margin-bottom-4">
         <strong>{{ transaction.sentAtFormatted }}</strong>
       </p>
       {% set transactionRows = [
           { key: { text: "Component" },  value: { text: transaction.componentId } },
-          { key: { text: "Reason" },     value: { text: transaction.interventionReason } }
+          { key: { text: "Reason" },     value: { text: transaction.interventionReason } },
+          { key: { text: "Code" },       value: { text: transaction.interventionCode if transaction.interventionCode else "N/A" } },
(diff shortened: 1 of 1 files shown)


src/frontend/views/user/_history.njk
```diff
@@ -48,6 +40,35 @@
...
     </div>
   {% endfor %}
+  <h2 class="govuk-heading-m govuk-!-margin-top-7">Field descriptions</h2>
+  {{ govukSummaryList({
+    rows: [
+      {
+        key: {
+          text: 'Originating Component'
+        },
+        value: {
```
edlouth, line 50: **Thought:** I don't know how helpful this is, but we could link to https://event-catalogue.internal.account.gov.uk/attributes/extensions-intervention-originating_component_id/ @ede-gds?

edlouth (commented):
Looks good, one small thought

src/frontend/views/user/_history.njk
ede-gds: Yeah, I think we should link to the event catalogue wherever we can (even though it's a bit rubbish). I might also ask Lara if she can give us a description for each of these fields which makes more sense to the fraud analysts but that shouldn't block this PR 👍

kirstyallen1: So for now do we want to leave the descriptions in and add a third column with the link to relevant area in the event catalogue, or just replace the description with the link to the specific page in the event catalogue for each field?

ede-gds: What about just a `(Source)` link at the end of the description text?

edlouth (approved):

PR #1139, opened 2026-07-27
FPAD-8463: AIS auth mop up

edlouth:
## What

AIS auth mop up

### Issue tracking

- [FPAD-8463](https://govukverify.atlassian.net/browse/FPAD-8463)

[FPAD-8463]: https://govukverify.atlassian.net/browse/FPAD-8463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/frontend/app.ts
+++ b/src/frontend/app.ts
@@ -20,7 +20,7 @@ import { Authoriser } from './authoriser';
 
 declare module 'fastify' {
   interface FastifyRequest {
-    awsLambda: { event: APIGatewayProxyEvent; context: Context };
+    awsLambda?: { event: APIGatewayProxyEvent; context: Context };
   }
 }
 
@@ -54,7 +54,7 @@ function formatDate(value: string | number): string {
 
 export const generateVerifyRequest =
   (authoriser: Authoriser) => async (request: FastifyRequest, reply: FastifyReply) => {
-    const authoriserResult = await authoriser.verify(request.awsLambda.event.requestContext.authorizer, request.url);
+    const authoriserResult = await authoriser.verify(request.awsLambda?.event.requestContext.authorizer, request.url);
 
     if (!authoriserResult.success) return reply.status(401);
   };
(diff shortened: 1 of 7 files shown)


src/frontend/app.ts
```diff
@@ -54,7 +54,7 @@ function formatDate(value: string | number): string {
 
 export const generateVerifyRequest =
   (authoriser: Authoriser) => async (request: FastifyRequest, reply: FastifyReply) => {
-    const authoriserResult = await authoriser.verify(request.awsLambda.event.requestContext.authorizer, request.url);
+    const authoriserResult = await authoriser.verify(request.awsLambda?.event.requestContext.authorizer, request.url);
```
ede-gds, line 57: **Question:** Is this definitely the only optionality in the `request` object? Based on the principle of not trusting anything at the boundaries is it worth parsing it with a little Zod schema?

src/handlers/frontend-handler.ts
```diff
@@ -16,7 +16,7 @@ const config = AppConfigService.getInstance().getConfigObject(['statusApiUrl', '
 
 const featureFlags = FeatureFlagsFromEnvironmentVariables.getInstance();
 
-const authoriser = new JwtAuthoriser();
+const authoriser: Authoriser = featureFlags.isEnabled('disableAuth') ? new StubAuthoriser() : new JwtAuthoriser();
```
ede-gds, line 19: 😍

src/infra/frontend/template.yaml
```diff
@@ -10,6 +10,7 @@ Metadata:
         - E3031
         - E3510
         - E3033
+        - W8003
```
ede-gds: **Thought:** We should probably have comments explaining what these are actually ignoring

ede-gds (commented):
lgtm, just one comment of substance

src/infra/frontend/template.yaml
edlouth: Done

src/frontend/app.ts
edlouth: I don't have a great answer to this, I could expand the optionality of the type, but this is somewhat getting into the inner workings of Lambda and to some extent it should throw if the object doesn't match the expected types

ede-gds: I think you're right, but one way of making it throw is to run it through a parser that spits this out - not a big deal though, and I actually meant to approve this PR anyway

ede-gds (approved):

ede-gds (approved):

PR #1141, opened 2026-07-28
Fpad 8632 move to app config

edlouth:
## What

Move `statusApiUrl` and `txmaQueueUrl` to app config

## Why

Single config object to inject app config

### Issue tracking

- [FPAD-8632](https://govukverify.atlassian.net/browse/FPAD-8632)

[FPAD-8632]: https://govukverify.atlassian.net/browse/FPAD-8632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/test/app-config-service.test.ts
+++ b/src/services/test/app-config-service.test.ts
@@ -69,6 +69,8 @@ describe('AppConfigService', () => {
   });
 
   it('should return the value of all environment variables', () => {
+    vi.stubEnv('TXMA_QUEUE_URL', 'https://sqs.eu-west-2.amazonaws.com/111122223333/TxMAQueue');
+
     const appConfig = AppConfigService.getInstance();
     expect(appConfig.awsRegion).toEqual('aws_region');
     expect(appConfig.tableName).toEqual('table_name');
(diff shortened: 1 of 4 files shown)


src/services/test/app-config-service.test.ts
```diff
@@ -69,6 +69,8 @@ describe('AppConfigService', () => {
   });
 
   it('should return the value of all environment variables', () => {
+    vi.stubEnv('TXMA_QUEUE_URL', 'https://sqs.eu-west-2.amazonaws.com/111122223333/TxMAQueue');
```
ede-gds: **Suggestion:** This test should probably test the value of the two new variables from this PR. Or maybe this test is just a stupid change detector and we can just get of it

ede-gds (approved):
lgtm

src/services/test/app-config-service.test.ts
edlouth: I think this test is a bit silly to be perfectly honest, we should move away from these custom getters where possible and use `getConfigObject`

ede-gds: 🪓

PR #1145, opened 2026-07-29
FPAD-8595: Improve intervention terminology

kirstyallen1:
## What
Improve intervention terminology as per user feedback:
- Non-suspension interventions which have been dealt with by the user should be COMPLETED (instead of MITIGATED)
- Suspension interventions which have been removed for any reason should use the verb UNSUSPENDED (instead of MITIGATED) - (mitigated is listed in the ticket but I think this should be instead of REMOVED?)

## Why
We want to use terminology that is familiar to our users 

### Issue tracking

- [FPAD-8595](https://govukverify.atlassian.net/browse/FPAD-8595)

[FPAD-8595]: https://govukverify.atlassian.net/browse/FPAD-8595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/frontend/app.ts
+++ b/src/frontend/app.ts
@@ -5,7 +5,13 @@ import formbody from '@fastify/formbody';
 import nunjucks from 'nunjucks';
 import path from 'node:path';
 import { existsSync } from 'node:fs';
-import { AccountHistory, HistoryLine, InterventionClientInterface } from '@govuk-one-login/ais-status-sdk';
+import {
+  AccountHistory,
+  HistoryLine,
+  InterventionClientInterface,
+  InterventionName,
+  InterventionState,
+} from '@govuk-one-login/ais-status-sdk';
 import { APIGatewayProxyEvent, Context } from 'aws-lambda';
 import { FeatureFlags } from '../services/feature-flags';
 import cookie from '@fastify/cookie';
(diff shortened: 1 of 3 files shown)


src/frontend/app.ts
```diff
@@ -218,10 +224,32 @@ export function init(
...
+      return 'COMPLETED';
+  }
+  if (line.interventionState === InterventionState.REMOVED &&
+    (line.interventionName === InterventionName.TEMPORARY_SUSPENSION ||
+      line.interventionName === InterventionName.PERMANENT_SUSPENSION)
+  ) {
+    console.log('unsuspended')
+    return 'UNSUSPENDED'
+  }
+  console.log('always here')
```
edlouth: I think this needs removing

edlouth (approved):
Just a couple of `console.log`s to remove, otherwise looks good.

src/frontend/app.ts
kirstyallen1: Thought I'd caught them all 🥲 Good spot - will remove this one

edlouth: Probably we should have a lint rule that block console output (which can then be overridden by a comment line)

PR #1146, opened 2026-07-31
FPAD-8589: Distinguish between automated and human applied interventions

kirstyallen1:
## What
Adds a tag element with appropriate text to each intervention shown on the history page to distinguish between which interventions in the history are automated, via FIA, or SIRA

## Why
Based on user feedback, the users would like to be able to differentiate which interventions have come from the CMS, FAI, or TICF.

### Issue tracking

- [FPAD-8589](https://govukverify.atlassian.net/browse/FPAD-8589)

[FPAD-8589]: https://govukverify.atlassian.net/browse/FPAD-8589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/frontend/views/user/_history.njk
```diff
@@ -5,6 +5,15 @@
...
       <p class="govuk-body-xs govuk-hint govuk-!-margin-bottom-4">
         <strong>{{ transaction.sentAtFormatted }}</strong>
+        {% if transaction.originatingComponent == "TICF CRI" %}
+          {% set tagDescription = "AUTOMATED" %}
+        {% elif transaction.originatingComponent == "FAI" %}
+          {% set tagDescription = "FAI" %}
+        {% elif transaction.originatingComponent == "CMS" %}
+          {% set tagDescription = "SIRA" %}
+        {% endif %}
+        <strong class="govuk-tag govuk-tag--blue govuk-!-margin-left-7"
```
edlouth: **Suggestion:** I would be tempted by a different colour for each status.

kirstyallen1: I have also been considering different colours per status so will add this change!

src/frontend/views/user/_history.njk
```diff
@@ -5,6 +5,18 @@
...
+        {% if transaction.originatingComponent == "TICF CRI" %}
+          {% set tagDescription = "AUTOMATED" %}
+          {% set tagColour = "purple" %}
+        {% elif transaction.originatingComponent == "FAI" %}
+          {% set tagDescription = "FAI" %}
+          {% set tagColour = "blue" %}
+        {% elif transaction.originatingComponent == "CMS" %}
+          {% set tagDescription = "SIRA" %}
+          {% set tagColour = "teal" %}
+        {% endif %}
```
ede-gds, line 20: **Suggestion:** Maybe `SIRA` and `FAI` should be the same colour? We could use teal / purple because they're quite different.

**Suggestion:** If `originatingComponent` isn't set the content will be empty, right? Maybe we should be explicit and have it say `SOURCE UNAVAILABLE` or something

ede-gds (approved):
looks sensible, a couple of minor things

src/frontend/views/user/_history.njk
kirstyallen1: Done :)

